Issue #4, opened 2019-12-02
Add support for typed statement cache support


bikeshedder:
This is currently blocked by https://github.com/sfackler/rust-postgres/pull/515

bikeshedder:
The PR request was merged but no new version of `tokio-postgres` released, yet. Once a new release is made this can be implemented.

PR #20, opened 2019-12-26
Remove call to `to_owned` in statement cache

walfie:
Avoids allocating a String on each call to `prepare` if it's already in
the cache. Also ran `rustfmt` which changed one line.

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -43,8 +43,8 @@ use futures::FutureExt;
 use log::{info, warn};
 use tokio::spawn;
 use tokio_postgres::{
-    tls::MakeTlsConnect, tls::TlsConnect, Client as PgClient, Config as PgConfig, Error, Socket,
-    Statement, Transaction as PgTransaction, types::Type
+    tls::MakeTlsConnect, tls::TlsConnect, types::Type, Client as PgClient, Config as PgConfig,
+    Error, Socket, Statement, Transaction as PgTransaction,
 };
 
 /// A type alias for using `deadpool::Pool` with `tokio_postgres`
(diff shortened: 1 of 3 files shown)


bikeshedder:
Now that the new `tokio-postgres` version is released I could finally fix #4: See 89c8297e69c6385792163b5dfd3fdbad03e6f7f6

I did implement that by changing `HashMap<String, Statement>` to `HashMap<(String, Vec[Type]), Statement>` which makes it impossible to use your optimization.

I could create two HashMaps. One for `prepare` and another one for `prepare_typed` but I'm unsure if that has any benefit over one extra string allocation.

walfie:
Hmm, I think it may be possible to keep one hashmap, but instead of using a tuple as the key, to use a custom struct, and then implement `std::borrow::Borrow` for it. I'll look into it, and update this PR if I can get it to work.

walfie:
Wasn't able to do the `std::borrow::Borrow` thing I thought of originally (due to [limitations](https://www.reddit.com/r/rust/comments/8u9cdz/the_borrow_trait_is_too_weak_and_requires_dynamic/)) but I was able to have a version using `Cow` to abstract over owned vs borrowed values.

bikeshedder:
I didn't know about this use for `Cow::Borrowed` and `Cow::Owned`. Thanks a lot for pointing that out. I just found this great blog article which explains the idea behind it: https://jwilm.io/blog/from-str-to-cow/

I just moved the actual logic into the `StatementCache` so that the `prepare_typed` methods could become a bit simpler: https://github.com/bikeshedder/deadpool/commit/8df4a43ab0455a317e57c7f40b3f8079dc82da63

[`deadpool-postgres 0.4.1`](https://crates.io/crates/deadpool-postgres/0.4.1) is now available on `crates.io`.

PR #22, opened 2019-12-27
impl DerefMut for ClientWrapper

walfie:
While making this change I realized this used to exist, but was removed per [the changelog](https://github.com/bikeshedder/deadpool/blob/master/postgres/CHANGELOG.md#v021):

> v0.2.1
> deadpool_postgres::Client no longer implements DerefMut which was not needed anyways.

It's not immediately clear why it was removed, but is there a suggested alternative way to get a mutable reference to the underlying client?

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -36,7 +36,7 @@
 #![warn(missing_docs)]
 
 use std::collections::HashMap;
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 use async_trait::async_trait;
 use futures::FutureExt;
@@ -168,6 +168,12 @@ impl Deref for ClientWrapper {
     type Target = PgClient;
     fn deref(&self) -> &PgClient {
         &self.client
+    }
+}
+
+impl DerefMut for ClientWrapper {
+    fn deref_mut(&mut self) -> &mut PgClient {
+        &mut self.client
     }
 }
 


bikeshedder:
Actually `DerefMut` should not be needed at all. The only function taking a `&mut self` is `transaction` which is wrapped so it supports the statement cache. Do you need access to a `&mut tokio_postgres::Client` in your code? It doesn't hurt to have that function but I didn't see any need for it when reading through the methods provided by `tokio_postgres::Client`.

walfie:
There are other libraries that are expecting a `&mut Client` (for example, I'm using [refinery](https://github.com/rust-db/refinery) for migrations).

Currently I'm working around it by creating a separate `Client` the normal way on app startup to get a `&mut Client`, running the migrations, and then throwing it away and making a pool and using that for the rest of the app, but it would be nice to be able to just use the pool.

bikeshedder:
I see. Personally I'd keep that connection for migrating and runtime pool separate from each other but that's really a matter of taste. It also makes it more compatible if `tokio-postgres` adds another method with `&mut self` in the future.

walfie:
Thanks! Yeah, I think overall it's not a huge deal to have to make a separate connection once on startup for this specific case.

I was mostly confused since I expected to be able to use a `ConnectionWrapper` interchangeably with a `Client` anywhere, and was wondering why it didn't work here, until doing a bunch of searching to find out that `DerefMut` wasn't implemented.

PR #24, opened 2020-01-11
Add explicit add function to pool

svenknobloch:
Add a method to allow for adding existing instances of a pool value to the pool. Useful when pooling a set of pre-initialized objects that can't be created as needed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -294,6 +294,29 @@ impl<T, E> Pool<T, E> {
         obj.state = ObjectState::Ready;
         Ok(obj)
     }
+    /// Add existing value to the pool. Returns whether or not the value was successfully added.
+    pub fn add(&self, value: T) -> bool {
+        let max_size = self.inner.config.max_size;
+        let mut current_size;
+        while {
+            // Get current size and check if max
+            current_size = self.inner.size.load(Ordering::Relaxed);
+            current_size < max_size
+        } {
+            // Try to increment size and push value
+            let old_size =
+                self.inner
+                    .size
+                    .compare_and_swap(current_size, current_size + 1, Ordering::Relaxed);
+            if old_size == current_size {
+                self.inner.queue.push(value).unwrap();
+                self.inner.available.fetch_add(1, Ordering::Relaxed);
+                return true;
+            }
+        }
+
+        false
+    }
     /// Retrieve status of the pool
     pub fn status(&self) -> Status {
         let size = self.inner.size.load(Ordering::Relaxed);
--- a/tests/explicit_add.rs
+++ b/tests/explicit_add.rs
@@ -1,0 +1,44 @@
+use async_trait::async_trait;
+
+use deadpool::{Pool, RecycleError, RecycleResult};
+
+struct Manager;
+
+#[async_trait]
+impl deadpool::Manager<(), ()> for Manager {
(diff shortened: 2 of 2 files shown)


bikeshedder:
I like the idea of it but I currently wonder about the actual use case. Do you need a pool that entirely consists of objects which are created upfront?

svenknobloch:
I'm planning on using it for command queues from gfx-hal/vulkan. When you initialize the gpu, you have to request the queues at the same time and then can't create any more without reinitializing. This way I can pool the queues and retrieve one whenever I need it.

src/lib.rs
```diff
@@ -294,6 +294,29 @@ impl<T, E> Pool<T, E> {
...
+        while {
+            // Get current size and check if max
+            current_size = self.inner.size.load(Ordering::Relaxed);
+            current_size < max_size
+        } {
+            // Try to increment size and push value
+            let old_size =
+                self.inner
+                    .size
+                    .compare_and_swap(current_size, current_size + 1, Ordering::Relaxed);
```
bikeshedder: While this makes sure that no two concurrent calls to `add` add more than `max_size` objects to the pool it is still possible to have a race between `add` and `get_timeout` which causes more objects being pushed to the queue than there are actual slots: https://github.com/bikeshedder/deadpool/blob/master/src/lib.rs#L280 - If there are `max_size` tasks currently in the critical section, but `size.fetch_add(1, Ordering::Relaxed)` has not been called yet you can end up adding an extra object to the queue and exceeding the `max_size` of the pool.

The current version of deadpool (0.4) relies entirely on the semaphore to ensure that there are no more than `max_size` tasks in that critical section. If the queue is empty it's safe to assume that `size` has not reached `max_size`, yet, and a new object can be created.

If it is really a viable use case to add existing object to a pool during runtime I'd change the `get_timeout` method to call `add` internally which uses the atomics to ensure that no more than `max_size` objects are ever added to the pool.

bikeshedder:
I see. Wouldn't it make more sense to add a special constructor method to the `Pool` which creates a pool instance with an already initialized set of objects, then?

Do you even need the Manager trait in your case? `create` is probably `unimplemented!()` and `recycle` is probably a no-op, too?

svenknobloch:
That would probably make more sense. Yeah the Manager is how you described it.

Would something along the lines of `from_existing(values: impl IntoIterator<Item=T>)` (with exact size iterator) and then some sort of default no-op manager implementation sound better to you? 

I originally wanted to just do the adding so that the pool initializers can remain since there are multiple (with/without the config). Adding the new initializer kind of conflicts with the `PoolConfig` since the max size can be derived from the iterator but the timeout config can't and adding an option with just a timeout will need updating if the config ever has another value other than just size and timeout.

bikeshedder:
I just started experimenting with a "`static`" version of the pool. It would provide a similar interface but leave out things like the `Manager` and backend errors as they can't occur. Actually I wonder if you even need timeouts or if this should be left to wrapping the `get` call with a `tokio::time::timeout()`.

svenknobloch:
That would work too. I do think that leaving the timeouts up to the user is simpler and makes more sense. It also doesn't really require much overhead and allows for varying timeouts for different use cases using the same pool.

In that case I'll close this pull request. Let me know if you need anything from me, I'd be happy to help implement some stuff.

Issue #25, opened 2020-01-11
Add an unmanaged version of the pool


bikeshedder:
@svenknobloch proposed in #24 to add an `add` function to the pool which allows using the pool where objects are not generated during runtime but upfront. This version of the pool does not need a manager and timeouts as no creation and/or recycling needs to happen.

bikeshedder:
I just added the implementation for that "passive" `Pool` to the repository. See https://github.com/bikeshedder/deadpool/blob/master/src/passive.rs

### Usage

```rust
use deadpool::passive::Pool;
#[tokio::main]
async fn main() {
    let pool = Pool::from_vec(vec![
        "foo".to_string(),
        "bar".to_string(),
    ]);
    {
        let s = pool.get().await;
        assert_eq!(s.len(), 3);
        assert_eq!(pool.status().available, 1)
    }
    assert_eq!(pool.status().available, 2);
}
```

The code is a lot simpler and the API of it is way better suited for fixed-size pools that don't require a manager to create/recycle/destroy objects as needed.

@svenknobloch Could you please check if this code is practical for your usecase?

All that's missing are some unit tests to ensure the correctness of it and it's ready to be pushed to `crates.io`.

If anyone has a suggestion for a better name I'm open for suggestions. My initial thought was `static` but having to write `r#static` in an public API didn't feel right. Other name candidates were `fixed` and `constant_sized` until I decided to put the bikeshedding aside and just call it `passive` for the time being.

svenknobloch:
Yep that works for me!

For the naming, I would name the module static internally and then exporting the name for the public api as `StaticPool` from the root, but that is of course up to you.

I also think it would make more sense to change `from_vec` to something like `from_iter` and have it take an `ExactSizeIterator` so that it's a bit more generic and can be used for more than just vecs. (I can PR this real quick if you want)

bikeshedder:
PRs are always welcome. 👍

Regarding naming things: I'm currently thinking of renaming the current "default" pool into `managed` and the other one into `unmanaged` as their main difference is the lack of a `Manager` for the unmanaged one. I would move the current pool into the `deadpool::managed` module then and keep a type alias in the crate root for backwards compatibility.

svenknobloch:
That seems like a fair solution to me.

bikeshedder:
This is now implemented and will soon be released as part of `deadpool 0.5`.

PR #26, opened 2020-01-11
Change `from_vec` to `from_exact_sized_iter`

svenknobloch:
Improve the `from_vec` method to be more generic. I was going to use the name `from_iter` but it might get confusing with the `std::iter::FromIterator` trait method. An alternative might just be to use `new` since there is no other constructor. #25

### DIFF
--- a/src/passive.rs
+++ b/src/passive.rs
@@ -9,7 +9,7 @@
 //!
 //! #[tokio::main]
 //! async fn main() {
-//!     let pool = Pool::from_vec(vec![
+//!     let pool = Pool::from_exact_sized_iter(vec![
 //!         "foo".to_string(),
 //!         "bar".to_string(),
 //!     ]);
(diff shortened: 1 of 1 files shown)


bikeshedder:
It's funny that you mention the `FromIterator` trait? How about implementing it? Looks to me like a perfect match for this kind of `Pool`. I'm also happy with the name `new` for this method as it's probably the only constructor for this struct that's ever going to be needed.

svenknobloch:
The `FromIterator` trait doesn't require the iter to implement `ExactSizeIterator` which means that the length is unknown and the `ArrayQueue` requires a known length (It might be possible to add an implementation of FromIterator to ArrayQueue but none currently exists). It would be possible to collect it into a vec and then do it but that would require allocation and at that point the user might as well do it themselves.

bikeshedder:
You're right. `FromIterator` doesn't really add anything.

I just tried adding the `FromIterator` and `IntoIterator` traits and realized that those two don't make an awful lot of sense. Unless someone needs this I think it's fine to have just the constructor function which accepts an `ExactSizeIterator` as argument.

PR #36, opened 2020-01-24
Make deadpool_postgres::config module public

tyranron:
This PR makes `deadpool_postgres::config` public.

## Motivation

At the moment there is quite friction when creating custom wrappers on top of `deadpool-postgres` crate. The problem is that types related to `Config` cannot be imported and used.

For example, [`ConfigError` returned by `Config::create_pool`](https://docs.rs/deadpool-postgres/0.5.2/deadpool_postgres/struct.Config.html#method.create_pool) cannot be used in custom signatures as is unreachable, so the only option left is just to unwrap it, which is not good.

The same with `config::SslMode`, `config::ChannelBinding` and others if you want to create custom mapping to those types.

## Checklist

- [x] `rustfmt`ed
- [x] `cargo test -p deadpool-postgres` pass and no warnings produced
- [x] `postgres/CHANGELOG.md` entry is added

### DIFF
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.5.3 (unreleased)
 
 * Add `#[derive(Clone)]` to `Config` struct
+* Make `config` module public
 
 ## v0.5.2
 
--- a/postgres/src/config.rs
+++ b/postgres/src/config.rs
@@ -1,3 +1,5 @@
+//! This module describes configuration used for [`Pool`] creation.
+
 use std::env;
 use std::fmt;
 use std::path::Path;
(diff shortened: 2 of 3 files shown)


bikeshedder:
Yeah. That is indeed an oversight by me. Originally I didn't even want this to be part of `deadpool` but then realized I had to wrap every single config type of `tokio-postgres` for this to work. It seams I forgot to make the `config` mod public afterwards. :man_facepalming:

bikeshedder:
I just released `deadpool-postgres 0.5.3` on `crates.io`: https://crates.io/crates/deadpool-postgres/0.5.3

Thanks for the PR. :+1:

PR #37, opened 2020-01-25
Allow nested transactions with deadpool_postgres::Transaction

tyranron:
This PR adds `DerefMut` implementation for `deadpool_postgres::Transaction` wrapper, so opens nested transactions capability supported by `tokio-postgres`:
- https://docs.rs/tokio-postgres/0.5.1/tokio_postgres/struct.Transaction.html#method.transaction
- https://docs.rs/tokio-postgres/0.5.1/src/tokio_postgres/transaction.rs.html#271-281

## Checklist
- [x] `rustfmt`ed
- [x] `postgres/CHANGELOG.md` entry is added

### DIFF
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## v0.5.4
+
+* Implement `DerefMut` for `Transaction` wrapper
+
 ## v0.5.3
 
 * Add `#[derive(Clone)]` to `Config` struct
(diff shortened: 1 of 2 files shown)


bikeshedder:
Actually the transaction wrapper was missing the `transaction` method so I added it: 0fa33c1b18005228e989b3471726203d17c590b6

Simply adding `DerefMut` would disable the statement cache for nested transaction. Adding `DerefMut` doesn't hurt however and increases the compatibility so I merged this PR.

I just released `deadpool-postgres 0.5.4` which includes those changes: https://crates.io/crates/deadpool-postgres/0.5.4

PR #60, opened 2020-07-15
deadpool-redis: Disable redis default features

alexlapa:
At the moment `deadpool-redis` imports `redis` with all default features enabled, and there a a lot of those:
1. With `default-features = false` its 25 dependencies and 20s compilation time.
2. With all default features enabled its 72 dependencies and 42s compilation time.

End users usually won't need all of the features, and should explicitly opt-in to use those they need. 

The only feature that `deadpool-redis` requires is `aio`, and `aio` requires async runtime. `redis` provides support for `tokio`, that can be enabled with `tokio-comp` feature, and `async-std` that can be enabled with `async-std-comp` feature. Both `tokio-comp` and `async-std-comp` features are default hence enabled in `deadpool-redis` right now.

In this PR i've made `tokio-comp` default one, since it is much more common atm. Other option might be making both `tokio-comp` and `async-std-comp` default.

This change can be considered kinda BC-ish.

### DIFF
--- a/redis/CHANGELOG.md
+++ b/redis/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## v0.?.?
 
 * Import `redis` with `default-features = false`.
+* Add `tokio-comp` default feature that enables [tokio](https://crates.io/crates/tokio) async runtime.
+* Add `async-std-comp` feature that enables [async-std](https://crates.io/crates/async-std) async runtime.
 
 ## v0.6.0
 
--- a/redis/CHANGELOG.md
+++ b/redis/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Change Log
 
-## v0.?.?
+## v0.7.0
 
-* Import `redis` with `default-features = false`.
-* Add `tokio-comp` default feature that enables [tokio](https://crates.io/crates/tokio) async runtime.
-* Add `async-std-comp` feature that enables [async-std](https://crates.io/crates/async-std) async runtime.
+* Import [redis](https://crates.io/crates/redis) with `default-features = false`.
+* Add `with-tokio` default feature that enables [tokio](https://crates.io/crates/tokio) async runtime.
+* Add `with-async-std` feature that enables [async-std](https://crates.io/crates/async-std) async runtime.
 
 ## v0.6.0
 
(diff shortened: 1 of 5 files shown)


alexlapa:
It seems that there are some issues with github, ive pushed more commits, but they are not appearing here.

Up to date diff can be found here: https://github.com/bikeshedder/deadpool/compare/master...alexlapa:disable-redis-default-features

Marking PR as draft until issues will be resolved.

bikeshedder:
I can see the other commits for this PR.

Regarding the `-comp` suffix. I think it's time to name the features like `tokio-postgres` does: https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/Cargo.toml#L30

- `with-tokio-0_2`
- `with-async-std-1`

I used to create an alias for the crates instead and leave the feature names without it. e.g. `config` is aliased as `config_crate` so it doesn't conflict with the `config` feature. While that works I guess that could be confusing for a lot of people looking at the sources and wondering what that mysterious `config_crate` is.

It's quite sad that features and dependencies are not namespaced.

bikeshedder:
btw. Since this is a breaking change the version of this crate must be bumped to `0.7.0`

alexlapa:
`*-comp` naming was taken from `redis`,  i agree that `with-*` looks better.

However, i'm not sure what you mean with version suffixes: `*-0_2`, `*-1`. These versions are pinned in `redis`'s `Cargo.toml`, and i'm not sure that we can or even should override them somehow.

bikeshedder:
You're right. Adding the version number probably doesn't make sense as it is a dependency of the redis crate.

bikeshedder:
After thinking about this for quite some time I decided not to "proxy" all the redis feature flags in this crate but rather make everyone aware that this crate uses the `redis` crate internally and just disable the default features of the redis crate by default.

The upcoming redis 0.18 version disables most of the default features anyways. For experiment sake and to honor the current tokio 0.3 release I tried disabling the default features in the `master` branch. You can give this a try. Be aware that it also uses the `tokio-compat-02` crate and does some experimental stuff. I hope redis 0.18 is soon released which makes this hackery obsolete.

See mitsuhiko/redis-rs#396

alexlapa:
> disable the default features of the redis crate by default

That will fix this issue. However, this makes `cd redis && cargo build` fail. No big deal, i guess. Closing this PR.

As for the tokio 0.3, i cant test it properly, cause have to stick to 0.2, there are many other dependencies that rely on that.

PR #61, opened 2020-08-05
Support rustls feature from lapin

mkpankov:
Signed-off-by: Michael Pankov <work@michaelpankov.com>

### DIFF
--- a/lapin/Cargo.toml
+++ b/lapin/Cargo.toml
@@ -32,3 +32,4 @@ dotenv = "0.15.0"
 [features]
 default = ["config"]
 config = ["config-crate", "serde", "deadpool/config"]
+rustls = ["lapin/rustls"]


bikeshedder:
It is impossible to publish crates to crates.io which depend on Git repositories. If you need `rustls` support there is little I can do right now. I'm marking this PR as blocked for that reason.

If you want to use a `lapin` version from Git you're best of using a fork of `deadpool-lapin` until a new `lapin` version is released to crates.io. If you want this to be part of the official `deadpool` repository I could provider a branch for that until a new lapin version is released.

mkpankov:
Sorry, turned out to be not necessary. Got confused with features.

bikeshedder:
No problems. btw. adding features to `deadpool-lapin` which are proxied directly to the `lapin` crate aren't necessarily a bad thing. This is also something #60 is about. It makes perfect sense to only depend on `deadpool-lapin` rather than depending on the `lapin` crate directly and possibly having a version mismatch between `deadpool-lapin` and `lapin`.

PR #62, opened 2020-08-10
Update redis dependency to v0.17

x04:
No reason not to be on the latest redis version.


bikeshedder:
Before merging this I want to check if it is possible to depend on `redis >= 0.15, <=0.17` this would mean that `deadpool-redis` just needs to increment its micro version and limit build breakages. This is related to #60 in a way. Tests are needed for this in order to figure out if the compatibility is given for all redis version combinations.

PR #69, opened 2020-11-13
Provide AsRef<T> / AsMut<T> impls for Object

kim:
It is usually preferable to use an AsRef bound for generic parameters,
as it is more general than Deref. Since Object is already Deref /
DerefMut, we can trivially provide corresponding AsRef / AsMut impls.

### DIFF
--- a/src/managed/mod.rs
+++ b/src/managed/mod.rs
@@ -149,6 +149,18 @@ impl<T, E> DerefMut for Object<T, E> {
     }
 }
 
+impl<T, E> AsRef<T> for Object<T, E> {
+    fn as_ref(&self) -> &T {
+        self
+    }
+}
+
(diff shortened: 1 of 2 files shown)


bikeshedder:
That makes a lot of sense. Thanks for the PR.

I do wonder however when you would want to write a function with a `AsRef<Connection>` parameter rather than `&Connection`.

Do you have any examples for that? Is this something worth mentioning in the docs maybe?

kim:
You are right that it is often possible to make the caller borrow -- it's notationally convenient though if it doesn't have to, just like the many examples of `AsRef<slice-ish type>` you'll find in `std`. I am also using trait objects (`dyn AsRef<Connection>`), where I don't want to be limited to smart pointers (ie. `Deref`) -- I can pass a much bigger, potentially stateful struct value without ceremony.

PR #70, opened 2020-11-21
Update README.md

kolektiv:
It looks like the small table of features/deps is wrong - looking at the examples (and then `Cargo.toml`) it looked like `async-trait` is only needed for managed, not un-managed. So this is a quick fix, presuming I'm correct on that...


bikeshedder:
That is indeed the case. Thanks a lot! :+1:

PR #71, opened 2020-11-22
Add a simple actix-web + deadpool-redis example

gyfis:
Hi, this PR adds a simple example of using `actix-web` + `deadpool-redis` together. I intentionally didn't include the config+env setup you have elsewhere, because 
- you already have it in the actix-web postgres example;
- I wanted to keep this example as simple as possible;
- the redis deployment I use expects `REDIS_URL`, and I wanted that to be explicit here as well.

Hope that's fine! Let me know if you'd like me to update the example. 

Thanks! Tomas

### DIFF
--- a/examples/redis-actix-web/.gitignore
+++ b/examples/redis-actix-web/.gitignore
@@ -1,0 +1,4 @@
+/target
+**/*.rs.bk
+Cargo.lock
+.env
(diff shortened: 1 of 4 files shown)


bikeshedder:
Looking good. Thanks a lot.

PR #85, opened 2021-02-15
Only use a default host when no others are given

### LINKED ISSUE #84: deadpool-postgres includes a second default host even when the "host" Config param is specified
See https://docs.rs/deadpool-postgres/0.7.0/src/deadpool_postgres/config.rs.html#328-345

If you specify the `host` parameter but not `hosts`, dedpool-redis tacks on a second host. On Unix, this will be `/tmp` if you aren't running a postgres server locally. As a workaround, you can instead use the `hosts` parameter instead with a single host.

This has the frustrating effect of hiding any errors from connecting to the actual remote host. If tokio-postgres fails to connect to the first host, it will move on and try the second one. Instead of your actual error, you'll get something like `error connecting to server: No such file or directory (os error 2)`. I think that's what was behind the error message in https://github.com/bikeshedder/deadpool/issues/82.

**Repro**
```rust
let mut cfg = deadpool_postgres::Config::new();
cfg.dbname = Some("mydb".into());
cfg.host = Some("my-remote-host.com".into());
println!("{:?}", cfg.get_pg_config().unwrap());
```

```
Config { ... host: [Tcp("my-remote-host.com"), Unix("/tmp")], ...}
```

nickfrostatx:
Fixes #84.

### DIFF
--- a/postgres/src/config.rs
+++ b/postgres/src/config.rs
@@ -332,7 +332,8 @@ impl Config {
             for host in hosts.iter() {
                 cfg.host(host.as_str());
             }
-        } else {
+        }
+        if self.host.is_none() && self.hosts.is_none() {
             // Systems that support it default to unix domain sockets
             #[cfg(unix)]
             {


bikeshedder:
Good catch!

That's indeed a terribly oversight by me. I remember pondering if I really want a `host` and `hosts` config option that do almost the same. Since having a single host is more common and it's a 1:1 mapping from the configuration file I decided that just having `hosts` would be highly confusing for everyone.

Issue #92, opened 2021-03-20
Incorrect Pool Status counts


pjenvey:
We're seeing managed Pool's `Status::available` count exceed its `max_size` (https://github.com/mozilla-services/syncstorage-rs/issues/803).

In `Pool::timeout_get`: `available`'s only decremented once at the beginning of the method but its loop potentially evicts > 1 idle connection.

1. Consider the case of `Manager::recycle` evicting "stale" connections from the Pool by returning `PoolError` and the loop encountering multiple errors. Idle connections are pop'd from the pool for every error, `available` (and `size`) should decrement for each. 
2. Errors propagated from `apply_timeout()...await?` might also cause incorrect stats.

pjenvey:
#93 is a potential fix for 1.

bikeshedder:
Whops. I referenced the wrong issue with my commit. This is fixed in 15f94710a7e653941924379ae66fb2371a6946ee

PR #93, opened 2021-03-20
fix Status counts when recycle evicts objects

pjenvey:
Issue: #92

### DIFF
--- a/src/managed/mod.rs
+++ b/src/managed/mod.rs
@@ -228,8 +228,6 @@ impl<T, E> Pool<T, E> {
     /// Retrieve object using a different timeout config than the one
     /// configured.
     pub async fn timeout_get(&self, timeouts: &Timeouts) -> Result<Object<T, E>, PoolError<E>> {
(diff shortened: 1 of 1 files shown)


src/managed/mod.rs
```diff
@@ -265,6 +263,7 @@ impl<T, E> Pool<T, E> {
 
         loop {
             obj.state = ObjectState::Receiving;
+            self.inner.available.fetch_sub(1, Ordering::Relaxed);
```
bikeshedder, line 266: When delaying this until the beginning of the loop the `available` counter doesn't get decreased for waiting tasks/workers. The original idea of `available` is that a negative number indicates that there are not enough connections available and there are `-available` waiting tasks/workers.

The only change needed is the other part of the patch you proposed. The count needs to be reduced by one if recycling fails. All other cases are already covered by the `Drop` implementation of the `Object`.

bikeshedder (commented):
I'm fixing this upstream right now.

Issue #88, opened 2021-02-19
Add feature to use async-std synchronization primitives instead of the ones from tokio


Fishrock123:
[`http-client`](https://github.com/http-rs/http-client) now depends on deadpool for keep-alive (connection pooling) support when using it's lightweight `async-h1` backing implementation.

That now pulls in `tokio`, which is a bit undesirable when using a different async runtime. It would nice to have feature flags for that, to allow better interoperability with options with the async Rust ecosystem.

bikeshedder:
The `tokio` dependency is only there for some sync primitives (`tokio::sync::Mutex` and `tokio::sync::Semaphore`) and `tokio::time`. The runtime is not used or depended on at all. Therefore you don't get the `tokio` runtime when using `deadpool`.

If you check the `Cargo.toml` of `deadpool` you will see that it only depends on the `sync` and `time` feature of tokio. Therefore only a very small subset of `tokio` is being used. The locking primitives of `tokio::sync` are **executor independent** and so is `deadpool`. I could replace those by something like [`async-lock`](https://crates.io/crates/async-lock). I prefer the `tokio` primitives as they are thoroughly tested and thanks to loom proven to be correct.

Tokio used to be split into various crates (e.g. [`tokio-sync`](https://crates.io/crates/tokio-sync)). Just think of `tokio` with the default features disabled and `sync` enabled as a 1:1 replacement of this former `tokio-sync` crate. The `tokio` sync primitives are in fact runtime independent.

The only issue I know about is the use of `tokio::time`. Timeouts currently only work with `tokio` (See #68). I was hesitant to add timeouts in the first place and soon after adding them realized that they render this feature to be runtime specific (See #9). This is something I want to take care of very soon. I was hoping for [`async_executors`](https://crates.io/crates/async_executors) to add timeout support (See najamelan/async_executors#3) but that hasn't happened, yet.

bikeshedder:
I was wondering if you were just worried about the compatibility with the `async-std` runtime or wanted to see `deadpool` support multiple `Mutex`/`RwLock`/`Semaphore` implementations to cut down the build size?

I added the `documentation` tag to this issue as I should add this to the FAQ: The use of `tokio::sync` does not mean `deadpool` depends on the `tokio` runtime features. The README does state that `deadpool` is compatible with any executor. The `tokio` dependency could deceive people into believing that is also depends on the `tokio` runtime which is not true unless using timeouts while #68 is not fixed.

[...]

PR #95, opened 2021-04-09
Make tokio an optional dependency for core module

sunng87:
At the moment, tokio is a required dependency even if we are using async-std. This patch make it optional to clean up dependency tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ async-trait = { version = "0.1.17", optional = true }
 config-crate = { package = "config", version = "0.11", default-features = false, optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 # async runtimes
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1", features = ["sync"], optional = true }
 async-std = { version = "1", optional = true }
 
 [dev-dependencies]


bikeshedder:
`tokio/sync` is a required dependency. (not tokio, as explained in #88)

`tokio` used to be split into `tokio-sync`, `tokio-util`, `tokio-io`, etc. Currently it's an all-in-one crate with feature flags. If you add `tokio = { version = "1", features = ["sync"] }` you are only getting `tokio::sync` and not the rest of it. The `tokio` runtime is NOT included.

`deadpool` needs `tokio::sync::Semaphore` to function. That dependency is not optional.

The CI tests do pass because they are run with `--all-features` and therefore `tokio` is enabled as well.

bikeshedder:
If you want to make `tokio/sync` optional you first need to create a new `Semaphore` implementation that supports the same features as `tokio::sync::Semaphore` does:

- fair
- low level (adding and removing permits without a guard)
- closing (used for shutdown purposes)

The semaphore implementation from `async-lock` (which is reexported by `async-std`) supports neither of those features. It's rather primitive to be honest.

The semaphore from `futures-intrusive` could be used, but that comes with even more dependencies.

`tokio::sync::Semaphore` is currently the only choice available that has all the required features.

bikeshedder:
I've added tests to the CI that should cover this. If you rebase your PR to the latest version of the `master` branch the checks should now be failing.

sunng87:
Understand. Thank you for clarification.

PR #99, opened 2021-06-01
Export CreatePoolError from deadpool-redis

dimfeld:
This type is currently not exported from the crate, making it difficult to use without wrapping in something like `anyhow`.

### DIFF
--- a/redis/src/config.rs
+++ b/redis/src/config.rs
@@ -125,9 +125,12 @@ impl redis::IntoConnectionInfo for ConnectionInfo {
     }
 }
 
+/// An error returned when pool creation fails.
 #[derive(Debug)]
 pub enum CreatePoolError {
+    /// The pool configuration contained invalid options.
     Config(String),
+    /// Redis returned an error while creating the pool.
     Redis(redis::RedisError),
 }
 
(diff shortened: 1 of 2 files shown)


bikeshedder (approved):
Thanks a lot. I wonder why the rust compiler does not complain about this. This is for sure a private type used in a public interface. :thinking:

bikeshedder:
I just released deadpool-redis 0.8.1 which includes this fix:

- https://crates.io/crates/deadpool-redis/0.8.1

Thanks a lot for reporting and fixing this.

dimfeld:
I looked into this a bit more, and it appears to be semi-intentional that rustc doesn't complain about it, except it was only really intended to be used for sealed traits, not structs.

There's an RFC to add a lint for this: https://github.com/petrochenkov/rfcs/blob/9901089b2467b0813456479ba03642e9f0ac0912/text/0000-type-privacy.md#lint-3-voldemort-types-its-reachable-but-i-cant-name-it

And the tracking issue: https://github.com/rust-lang/rust/issues/48054.

BTW thanks for your work on deadpool. I had to make a custom Manager for my project and it was really easy :)

Issue #108, opened 2021-07-17
Customizing Connections


SergioBenitez:
For some databases, certain configuration parameters must be set per-connection. The canonical example is SQLite with its `PRAGMA`-based configuration. With respect to `deadpool`, the ideal place to do this would be immediately before/after a connection is made to be placed in the pool. As far as I can tell, `deadpool` provides no such hook to accomplish this, short of writing a custom `Manager`. Is that correct?

For reference, the analog in `r2d2` is [`connection_customizer`](https://docs.rs/r2d2/0.8.9/r2d2/struct.Builder.html#method.connection_customizer).

bikeshedder:
So far I've been trying to put as much configuration logic into the `Manager` implementations as possible.

I just wonder if that customizer needs to be `sync` or `async`... or maybe both? :thinking: 

What is the typical use case for such customizer? Does it issue DB queries or just change some options in the driver which are always non-blocking operations?

I am currently thinking of creating something like that:

```rust
/// An object modifier is called for every new connection that is
/// created by the manager.
pub enum ObjectCustomizer<T, W=T> {
    /// Use this variant if the function you are passing is know
    /// to never block. The function will be run as is for every
    /// new connection.
    NonBlocking(fn(obj: T) -> W),
    /// Use this variant if the given code is known to block.
    /// The function is wrapped inside [Runtime::spawn_blocking]
    /// making sure that it runs on a separate thread and doesn't
    /// block the async runtime.
    Sync(fn(obj: T) -> W),
    /// Use this variant if the given code is using async-await.
    Async(fn(obj: T) -> Pin<Box<dyn Future<Output=W> + 'static + Send>>),
}
```

I've been thinking about this for a bit now and quite like the idea of creating wrappers for the Manager instead of adding a new feature to the `Pool` implementation. e.g.

```rust
struct ManagerWrapper<T, W> {
    manager: Box<dyn Manager<T>>,
    customizer: ObjectCustomizer<T, W>,
}
```

Implementing a wrapping manager is no fun and way too many lines for a simple connection modification so I think it should be possible to use it like that:

```rust
let customizer = ObjectCustomizer::NonBlocking(|c| { c.set_some_parameter(); c });
let manager = Manager::new(...);
let wrapped_manager = customizer.wrap(manager);
let pool = Pool::new(wrapped_manager, 16);
...
```

What do you think of this approach?

bikeshedder:
@SergioBenitez What do you think of #109?

[...]

PR #109, opened 2021-07-17
Add experimental ObjectCustomizer code

bikeshedder:


bikeshedder:
There is one thing that I don't like about this approach. Every call to `create` or `recycle` creates a boxed future. While boxing is "cheap" it should not be needed when using the `ObjectCustomizer::NonBlocking` variant.

The extra box on `recycle` can be avoided when removing that `W` type and changing the `WrappedManager` implementation to:

```rust
impl<M> Manager for WrappedManager<M>
where
    M: Manager,
{
    type Type = M::Type;
    type Error = M::Error;
    ...
}
```

This would allow the implementation of `recycle` to avoid that extra allocation:

```rust
fn recycle<'life0, 'life1, 'async_trait>(
    &'life0 self,
    obj: &'life1 mut Self::Type
) -> Pin<Box<dyn Future<Output = RecycleResult<Self::Error>> + Send + 'async_trait>>
where
    'life0: 'async_trait,
    'life1: 'async_trait,
    Self: 'async_trait
{
    self.manager.recycle(obj)
}
```

My initial idea behind this was to allow changing the type so that I can remove the `W: From<Object<M>` from the pool but just realized that this is nonsense as the entire purpose of this is to let the `Pool` return a new type so traits can be implemented. e.g. `deadpool-wrapper` uses this to implement `ConnectionLike` so that users don't need to dereference the `Object<T>` whenever using it. The entire reason for being able to transform/wrap the type via `ObjectCustomizer` and `WrappedManager` is questionable at best.

Next I'll look into adding this feature to the `Pool` which would allow non-blocking modifications to be applied without any extra allocations.

bikeshedder:
Closing this in favor of a `PoolBuilder` and a `ObjectCustomizer` that can be added to the pool as described in #108

PR #115, opened 2021-07-18
Run drop of rusqlite::Connection inside spawn_blocking and fix panic handling

### LINKED ISSUE #113: User code can lead to infinite panics in `interact` in `deadpool-sqlite`
Here's the `interact` method, for reference:

https://github.com/bikeshedder/deadpool/blob/2c7a0b7aee915e775f8d751e0a73ec884bae6f71/sqlite/src/lib.rs#L136-L146

If `f()` panics, it will do so while the lock is held. This poisons the lock. Every attempt to `interact()` thereafter will `panic!()` because of the `unwrap()` on line 144 (itself because of the `unwrap()` in 142). I would suggest the following changes:
  1. Use a non-poisoning lock, like `parking_lot::Mutex()`.
  2. Propagate  errors from `spawn_blocking()`, don't panic.

Alternatively, detect the error and close the connection so that the poisoned mutex is never reused.

### LINKED ISSUE #114: `deadpool-sqlite` `ConnectionWrapper` is not `async`-safe
`ConnectionWrapper` wraps a `rustqlite::Connection` which implements a `Drop` that [may perform synchronous I/O](https://docs.rs/rusqlite/0.25.3/src/rusqlite/lib.rs.html#332-337):

```rust
impl Drop for Connection {
    #[inline]
    fn drop(&mut self) {
        self.flush_prepared_statement_cache();
    }
}
```

Thus, when (the last of the clones of) `ConnectionWrapper` is dropped, synchronous I/O will be performed in what is surely to be an async context. This is, of course, logically incorrect behavior that can lead to quite a few problems including panics.

In general, wrapping a synchronous connection _correctly_ (or any other I/O object) is incredibly error-prone. Please see our implementation in `rocket_sync_db_pools` where we've painstakingly tried to address all of the relevant issues: https://github.com/SergioBenitez/Rocket/blob/693f4f9ee50057fc735e6e7037e6dee5b485ba10/contrib/sync_db_pools/lib/src/connection.rs.

bikeshedder:
This fixes #113 and #114

### DIFF
--- a/sqlite/src/lib.rs
+++ b/sqlite/src/lib.rs
@@ -155,7 +155,10 @@ impl Drop for ConnectionWrapper {
         // Drop the `rusqlite::Connection` inside a `spawn_blocking`
         // as the `drop` function of it can block.
         tokio::task::spawn_blocking(move || {
-            arc.lock().unwrap().take();
+            match arc.lock() {
+                Ok(mut guard) => guard.take(),
+                Err(e) => e.into_inner().take()
+            }
         });
     }
 }
--- a/sqlite/src/lib.rs
+++ b/sqlite/src/lib.rs
@@ -23,14 +23,17 @@
 //! async fn main() {
 //!     let mut cfg = Config::new("db.sqlite3");
 //!     let pool = cfg.create_pool();
-//!     for i in 1..10 {
-//!         let mut conn = pool.get().await.unwrap();
-//!         let value: i32 = conn.interact(move |conn| {
-//!             let mut stmt = conn.prepare_cached("SELECT 1 + $1").unwrap();
-//!             stmt.query_row([&i], |row| row.get(0)).unwrap()
-//!         }).await;
-//!         assert_eq!(value, i + 1);
-//!     }
+//!     let conn = pool.get().await.unwrap();
+//!     let result: i64 = conn
+//!         .interact(|conn| {
+//!             let mut stmt = conn.prepare("SELECT 1")?;
+//!             let mut rows = stmt.query([])?;
+//!             let row = rows.next()?.unwrap();
+//!             row.get(0)
+//!         })
+//!         .await
+//!         .unwrap();
+//!     assert_eq!(result, 1);
 //! }
 //! ```
 //!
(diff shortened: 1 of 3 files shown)


sqlite/src/lib.rs
```diff
@@ -126,21 +160,46 @@ impl ConnectionWrapper {
...
+        })
+        .await
+        .map_err(|e| {
+            if e.is_panic() {
+                InteractError::Panic(e.into_panic())
+            } else if e.is_cancelled() {
+                InteractError::Aborted
+            } else {
+                unreachable!();
+            }
```
SergioBenitez, line 189: This seems unnecessarily error-prone. Why do this when you can write the panic-safe:

```rust

            if e.is_panic() {
                InteractError::Panic(e.into_panic())
            } else {
                InteractError::Aborted
            }
```

sqlite/src/lib.rs
```diff
@@ -126,21 +160,46 @@ impl ConnectionWrapper {
...
+
+impl Drop for ConnectionWrapper {
+    fn drop(&mut self) {
+        let arc = self.conn.clone();
+        // Drop the `rusqlite::Connection` inside a `spawn_blocking`
+        // as the `drop` function of it can block.
+        tokio::task::spawn_blocking(move || match arc.lock() {
+            Ok(mut guard) => guard.take(),
+            Err(e) => e.into_inner().take(),
+        });
```
SergioBenitez, line 203: I would add type annotations and explicit `drop`s to make it absolutely clear that you're dropping the connection.

sqlite/src/lib.rs
bikeshedder: I added that last `else` block if there is ever the chance that a third option is added.

But then... this is unlikely ... yet it would generate a meaningful panic. :thinking:

sqlite/src/lib.rs
bikeshedder: Right. `drop(...)` makes it absolutely clear. I'm currently rewriting this as a generic `SyncWrapper` which can be used for `deadpool-diesel` as well.

sqlite/src/lib.rs
SergioBenitez: Sure, but you would be making your dependent's code panic without 1) documentation and 2) it being a necessity. If you really want to find out if this happens, you _could_ insert a `debug_assert!` instead with a useful message to the user if it is reached. Then at least the panic won't happen in production. But even then, I see little to no value as the no-panic alternative is logically correct.

PR #117, opened 2021-08-03
Added a constructor for the redis config that only uses the url

BartWillems:
This makes creating the config slightly easier if you only want to use a
URL. This is because either the url or the connection field should be
`Some`. Otherwise you would have to create the struct yourself, by
setting all fields; or using the default implementation which requires
you to set the connection field to None.

### DIFF
--- a/redis/src/config.rs
+++ b/redis/src/config.rs
@@ -217,6 +217,15 @@ impl Config {
     pub fn get_pool_config(&self) -> PoolConfig {
         self.pool.clone().unwrap_or_default()
     }
+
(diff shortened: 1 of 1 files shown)


BartWillems:
I don't know if `from_url` is a good name for this, since it makes it sound as if you need `url::Url`.

bikeshedder:
Looks good to me.

I'm in process of introducing a builder pattern for the pool (and maybe configuration structs).

I'm currently out of office and will be looking into this in a few days when I'm back home. :+1:

PR #127, opened 2021-08-19
Replace "config" feature with "serde" feature

### LINKED ISSUE #126: Remove config feature and crate dependency
I just realized all deadpool crates depend on the `config` crate for no reason. Since the deprecation of the `from_env` methods the `config` crate is not used directly anymore.

The only place where the `config` crate is being used is in the tests and the examples.

Solution:
- Move `config` crate to `dev-dependencies`
- Rename `config` feature to `serde`

tyranron:
Fixes #126

Additionally, polishes implemenation, hardens some lints, and reworks CI to check all features combinations.

Despite this PR has a lot of changes, they're mostly cosmetic ones, and it doesn't change the implementation, except:
- replaces `config` feature with `serde` and disables `serde` by default;
- adds more `Debug` imps;
- adds `std::error::Error::source` impls whenever is possible;
- adds `#[must_use]` attr to some definions where it makes sense (builders, objects, etc).


bikeshedder (approved):
Wow! Amazing effort! :+1: 

There are just two things I'm curious about:
- Why did you add `#[must_use]` to the `new` methods?
- Why did you add `let _ = ...` to the `fetch_add` calls?

I'm going to merge this regardless of those two questions.

bikeshedder:
This PR is marked as `draft`. Do you plan on refactoring it into multiple smaller commits or are you planning to take a look at the rest of the codebase as well?

btw. `#![doc = include_str!("../README.md")]` requires a quite recent rust version, doesn't it? I'm not too worried about backwards incompatibilities as long as it is rust stable. Does this also run the tests which are part of the `README.md` file?

tyranron:
@bikeshedder 

> are you planning to take a look at the rest of the codebase as well?

Yes.

> btw. #![doc = include_str!("../README.md")] requires a quite recent rust version, doesn't it? I'm not too worried about backwards incompatibilities as long as it is rust stable.

I agree here. It's to much fuss maintaining duplicated READMEs.

> Does this also run the tests which are part of the README.md file?

Yes.

> - Why did you add `#[must_use]` to the new methods?

That's a habbit from using `clippy::pedantic`. Do you mind to turn `-D clippy::pedantic` on CI?

> - Why did you add `let _ = ...` to the `fetch_add` calls?

This is due to `unused_result` rustc lint. To mark places where a value/result is explicitly/intentionally ignored. Again, it's a little bit pedantic, but from my prespective is good for library code which could be oftenly read by others.

bikeshedder:
> I agree here. It's to much fuss maintaining duplicated READMEs. We also will have check for MSRV (minimum supported rust version) and minimal dependecies versions on CI.

That's why I added `sync-readme.py` to the codebase. It's synchronizes the content of `README.md` into the `lib.rs`. I created this script as `#![doc = include_str!("../README.md")]` didn't exist back then. I quite like the macro approach without real code duplication much better.

> This is due to unused_result rustc lint. (...)

I see. Pedantic is good. :+1: 

It looks a bit weird for those `fetch_add` cases, but I can live with that. :grinning:

tyranron:
@bikeshedder I have one question related to `deadpool-diesel`:   
There is a [`diesel/src/connection.rs` module](https://github.com/bikeshedder/deadpool/blob/b10981e6be636fa1f0974adfb13dea3adf1ecb10/diesel/src/connection.rs), but it's not really used and represents a dead code at the moment. While the used `Connection` is defined in `diesel/src/lib.rs`:
```rust
pub type Connection<C> = deadpool::managed::sync::SyncWrapper<C, Error>;
```
Is it OK to remove that dead code? Should it be eventually revived? Or for some reason just live it as it is?

bikeshedder:
This is indeed dead code. This is a leftover from the first implementation when `SyncWrapper` did not exist, yet. See #115

`deadpool-sqlite` and `deadpool-diesel` now use this synchronization wrapper which makes implementing pools for non-async code much simpler.

tyranron:
Almost there!

@bikeshedder what do you think about renaming `serde` feature to `serde_1` in the same manner as `tokio`/`async-std` related features named?  
It will make renamings less messy, some examples more cleaner, and somewhat pave the path for `serde 2.0` to not be a breaking change.

bikeshedder:
I've been thinking about something similar.

I find it annoying that `deadpool-redis` requires a new breaking release every time redis is updated. I'd prefer if it supported multiple redis versions instead. Prior to `redis 0.21` no actual code change was required to support the new redis version and my experiment of specifying a redis version range `>=0.18,<=20` didn't end well when people complained that it broke their builds due to `deadpool-redis` depending on a newer version after upgrading.

Adding a `redis_019`, `redis_020` and `redis_021` features those would either need to be mutally exclusive (which is inherently bad) or users would need to import the proper version in code. Something like that:

```rust
use deadpool_redis::redis019::Pool;`
```

That would probably be the cleanest solution, remove some coupling of the `redis` and `deadpool-redis` version number. Users of `deadpool-redis` would now need to upgrade their code and not just the `Cargo.toml` when upgrading. This could be solved with some aliases in code... but honestly... I don't really like that idea a lot.

I would love to tell cargo somehow that `deadpool-redis 0.9.1-redis019` is compatible with `deadpool-redis 0.9.2-redis019` but not compatible with `deadpool-redis 0.9.2-redis020`.

The only way this could be solved right now is by naming the package `deadpool-redis019` instead. Something I'm almost considering given the mess those versions create.

---

That's why I hope to have a API stable 1.0 release of `deadpool` by the end of this year. Then releases of `deadpool-FOO` will only need to be made if the backend package is updated with a new breaking version number.

So all this might end up being a non issue... :sweat_smile: 

---

Regarding serde. Calling it `serde_1` doesn't hurt but feels a bit unnecessary given the extreme API stability `serde` has shown in the past.

Regarding the feature names in general. `tokio-postgres` tends to call all features that enable a specific crate `with-<crate_name>-<version>`. For some reason I opted for `rt-...` for the runtime features. A decision I'm not so 100% sure about anymore especially as people tripped over it in the past assuming that it enables the given runtime automatically while it actually just enables the support for it. It still needs to be enabled in code.

So maybe calling all features that enable support for a specific dependency should be called `with-...` just like tokio-postgres does it?

bikeshedder:
btw. if all features were renamed to the `with-<crate_name>-<crate_version>` format I'd opt for `with-serde-1` just to make it look uniform. :wink:

tyranron:
@bikeshedder thanks for patience.

Regarding the feature naming I agree that keeping confusion/complexity inside the library is better than expose it to its users. Plus, I never liked the `with-<crate_name>-<crate_version>` format used by `tokio-postgres`. That's why I've left the feature name as `serde` and tried to beatify doc examples asap to hide possible weirdness.

Additionally:
1. I've revived `examples/` in Cargo `[workspace]` marking them as `publish = false` to avoid accidential publishing. Now those will be checked on CI to compile as well in `clippy` CI job (it checks the whole Cargo workspace), so no outdating should appear in future.
2. Some unused dependencies like `log` and `tokio` were removed from integration crates here and there. Also, the `futures` dependecy of `deadpool-postgres` was only used for `.then()` combinator which is trivially replaceable by `.await`, so I did that replacing.
3. Regarding MSRV check I've disabled it temporary for `deadpool-diesel` and `deadpool-lapin` as there is some mess happening in their transitive dependencies which cannot be solved naturally from our side (and shouldn't, though).

bikeshedder:
Just wanted to give you a short heads up.

I haven't found the time to review the entire changes, yet. I plan to merge this by this weekend. I've some other changes in my own pipeline that I held back until this is merged so it's definitely a priority to me.

Looking forward to the much cleaner code base, better checks, etc. p.p. :+1:

tyranron:
ping @bikeshedder

bikeshedder (approved):

bikeshedder:
Could you please rebase this branch once so it applies cleanly.

That one merge from `master` in your branch seams to be the problem here.

Ah, and while you're at it. `deadpool-redis 0.9` was released. That code lives in the `deadpool-redis-v0.9` branch and contains some cherry picked changes. Please update the `deadpool-redis` `CHANGELOG` accordingly.

bikeshedder:
I just realized I can commit to your branch, and even force push. I just rebased the code myself. :+1: 

Once the CI completed without errors I'll merge it. :crab:

Issue #121, opened 2021-08-18
Support for connection validation


trungda:
Do we plan to have support for something similar to `keepaliveTime` in HikariCP? This is to periodically validate the connections in the pool. The connections might no longer be valid if the db backend is restarted.

bikeshedder:
Most users don't need a active connection checking in the pool. They are cycled in a FIFO way in order to make sure everything is kept as "fresh" as possible.

Though there are two scenarios that I can think of where active connection refreshing can make a difference:

1. A service that has very little load (just a few requests per hour) which is still required to answer extremely fast. Thus all connections in the pool need to be in working condition and reconnecting hurts.
2. A service that handles less parallel requests than the available DB connections in the pool. In this case a small optimization would be to recycle the connections ahead of time and keep them available.

When writing deadpool I found that I want the guarantee of a "working" connection to be given as late as possible. I'd rather not hand out a connection that was checked 10s ago and considered to be in working condition which is dead by now. That's why I decided against implementing a clever background worker that removes dead connections, recycles connections ahead of time, etc.

In the end it resulted in a much simpler code base. Hence the name `deadpool` which I consider to be `dead simple` implementation wise.

---

That being said I've been thinking of adding some plugin/introspection API to `deadpool` so it's possible to add 3rd party code to work with the internal state of the pool. The actual objects live in a `std::sync::Mutex<VecDeque<_>>`. It's slightly dangerous to work with that directly as locking it without decrementing the semaphore first can lock up your executor.

A way to enable a 3rd party code to remove dead connections ahead of time could be a `get_pooled` method that returns all objects currently waiting in the pool. So for very simple health check that runs every minute you could write code like that:

```rust
const GC_TIMEOUT: usize = 60_000;
use tokio::time::{sleep, Duration};

fn spawn_pool_gc(pool: Pool) {
    tokio::spawn(async move {
        sleep(Duration::from_millis(GC_TIMEOUT)).await;
        for obj in pool.get_pooled() {
            if !some_clever_alive_check(obj) {
                Object::take(obj);
            }
        }
    });
}
```

This is a rather pragmatic approach and doesn't do any `ahead of time` recycling, adaptive pool resizing, etc.

---

Btw. you just gave me the idea for #123 which might provide very useful in this case, too.

Given a `creation_timestamp` you could remove objects older than `<duration>` from the pool.

trungda:
Thank you for the answer! Where do we define the `get_pooled` method?

[...]

PR #131, opened 2021-09-09
Feature/get pooled

bikeshedder:


bikeshedder:
This closes #121

bikeshedder:
While this implementation works as expected it's kinda silly. It locks the mutex once to retrieve the objects but locks the mutex once per object returned. It also updates the `available` counts individually which creates some silly statistics when used in a clean up task.

A better approach would probably to provide access to the underlying `VecDeque<M::Type>>` with a custom guard that takes care of updating the `available` counter afterwards. Code using this method should not contain any long running code anyways so locking the mutex for the duration of the clean up is probably a non-issue.

bikeshedder:
I'm no longer convinced that this feature is even worth having. The only use case of it has been replaced by `Pool::retain`, I think. Therefore I'm closing this PR unmerged.

PR #134, opened 2021-09-20
Give the `interact` closure a mutable reference

dchenk:
The major upcoming Diesel release (no ETA yet, I think) will require a mutable connection to interact with the database: https://github.com/diesel-rs/diesel/commit/75c688c3b246295f6f7182e0fec1b58cc685b4ed

As far as I can tell, the change proposed here shouldn't break existing consumers of `deadpool`.


bikeshedder (approved):
This was indeed an oversight. :+1:

PR #135, opened 2021-09-20
Add an `inner_obj` method to SyncWrapper

dchenk:
In some cases, it may make sense for a consumer of `deadpool_diesel` to independently ensure that all blocking IO work is done on a thread that's meant for such work. See https://github.com/dchenk/deadpool/commit/084e8e9d1dd036165085c8fdbb800deec12f34db for context. This PR adds an `inner_obj` method to `SyncWrapper` to allow consumers to get access to the internal `Arc<Mutex<Option<T>>>` and includes a doc comment explaining that anything you do with the object should be wrapped in a `spawn_blocking` closure.

### DIFF
--- a/src/managed/sync.rs
+++ b/src/managed/sync.rs
@@ -122,6 +122,14 @@ where
             .map_err(InteractError::Backend)
     }
 
+    /// Get the underlying object wrapped in a Mutex that's wrapped in an Arc.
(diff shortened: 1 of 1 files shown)


bikeshedder:
Closed in favor of #136

PR #136, opened 2021-09-20
Add SyncGuard for direct access to the inner object of SyncWrapper

bikeshedder:
@dchenk what do you think of this instead of #135 

Maybe `lock`/`try_lock` is a bit misleading? Though it makes it pretty explicit what's happening under the hood... :thinking:

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -36,7 +36,7 @@ pub mod sqlite;
 use deadpool::managed;
 
 pub use deadpool::{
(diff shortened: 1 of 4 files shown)


dchenk:
I like how this abstracts away from the internal `Option`. This is more complete than #135 — thank you. Feel free to close my PR if you're going to merge this.

dchenk:
A PR targeting this PR's branch: https://github.com/bikeshedder/deadpool/pull/138

PR #138, opened 2021-09-22
Make SyncGuard public

dchenk:
With the `SyncGuard` type private, it was a bit awkward to express the return type of `SyncWrapper::lock`.


bikeshedder:
That's indeed an oversight. It also needs to be added to `deadpool-sqlite` as well.

Maybe I should add some kind of a `reexports` module that can be used by pool implementations. A module containing everything that should be reexported by a pool implementation. SImilar to a prelude file.

PR #143, opened 2021-09-25
Replace ObjectState by DropGuard

### LINKED ISSUE #142: Simplify pool implementation by using a DropGuard rather than the ObjectState
The `ObjectState` has split the modification and undo operation on future cancellation into two separate code fragments. This is hard to reason about and has led to subtle bugs when implementing new features (e.g. #140).

The `ObjectState` can be completely removed and replaced by a very simple `DropGuard` implementation such as:

```rust
/// This structure calls a function/closure when it is dropped.
/// The [`DropGuard::disarm`] method stops this from happening.
pub(crate) struct DropGuard<F: Fn()>(pub(crate) F);

impl<F: Fn()> DropGuard<F> {
    pub(crate) fn disarm(self) {
        std::mem::forget(self)
    }
}

impl<F> Drop for DropGuard<F>
where
    F: Fn(),
{
    fn drop(&mut self) {
        (self.0)()
    }
}
```

bikeshedder:
This closes #142

### DIFF
--- a/src/managed/dropguard.rs
+++ b/src/managed/dropguard.rs
@@ -1,0 +1,47 @@
+/// This structure calls a function/closure when it is dropped.
+/// The [`DropGuard::disarm`] method stops this from happening.
+pub(crate) struct DropGuard<F: Fn()>(pub(crate) F);
+
+impl<F: Fn()> DropGuard<F> {
+    pub(crate) fn disarm(self) {
+        std::mem::forget(self)
+    }
+}
+
+impl<F> Drop for DropGuard<F>
+where
+    F: Fn(),
+{
+    fn drop(&mut self) {
+        (self.0)()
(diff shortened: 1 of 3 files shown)


bikeshedder:
@tyranron It would be nice if you could review this change since you are the one who knows the code best besides myself.

I'm going to merge this now as I have some other features that I'm working on that depend on that PR. It'd be nice to get some feedback if it actually made the implementation easier to reason about and less error prone.

tyranron (commented):
@bikeshedder seems to be a nice change 👍
It does, actually, simlifies code understanding and reasoning, from my point of view. And I cannot spot any misdirections atm.

bikeshedder:
Thanks a lot. :smile: :+1:

Issue #94, opened 2021-03-22
ArangoDB Support


ManevilleF:
Hello,

I'm the creator of the [aragog](https://gitlab.com/qonfucius/aragog) crate, an ODM/OGM for ArangoDB and I want to implement the `deadpool` pattern for my lib.

What is the preferred way?
- Add the implementation on this repository
- Add the implementation on a custom crate which can be referenced on your `README`
- Add an optional implementation directly on my crate with a feature gate

bikeshedder:
Having the implementation inside the repository would be nice as it would allow me to make releases for all implementation crates in one go whenever changing something from the `deadpool` crate (0.8 is incoming).

But then I wonder if I should rather extract the `Manager` trait into a separate crate which can be 100% stable and reduce the version dependencies even more. See #90

ManevilleF:
I don't know if it's a good idea to have everything inside of the `deadpool` repository, why don't you do something more similar to `r2d2` ?

Weasy666:
I have implemented `deadpool` for a `ArangoDB` backend, [`deadpool-arangodb`](https://github.com/Weasy666/deadpool-arangodb), and wanted to ask how i should proceed with publishing? I don't want to just use the `deadpool-` "namespace" without asking. So...can i publish it as `deadpool-arangodb` or should i use a different name? I am also open to moving the code inside the `deadpool` repository.

PR #153, opened 2021-10-07
[WIP] Add deadpool-arangodb

Weasy666:
This adds [`deadpool-arangodb`](https://github.com/Weasy666/deadpool-arangodb) to the repository.

I have marked the PR as `WIP`, because there is still a problem with `use serde_1::Deserialize;` being marked as `unresolved import` in `tests/arangodb.rs` and i can't figure out why.
I added `BuildError::Config` back to `deadpool`, because `arangors` does not use a config struct and as such does not have a config error that i could use for the case the pool was configured wrongly. If you prefer, we can leave `BuildError::Config` out and i can try adding such a value upstream (in arangors).
Another point is, i had to use the newest `arangors` rev from git, to avoid a build error in the release crate version, coming from the `typed-builder` dependency, which afaik published a breaking change without adhering to semver.

Addresses #94

### DIFF
--- a/arangodb/Cargo.toml
+++ b/arangodb/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "deadpool-arangodb"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 resolver = "2"
 authors = ["Daniel Wiesenberg <weasy@hotmail.de>"]
 description = "Dead simple async pool for ArangoDB"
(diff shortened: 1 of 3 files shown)


bikeshedder:
When running the tests you had to add `--features serde` since you depend on serde in the tests.

I just fixed that by making the `serde` feature non-optional. Since you are using `serde` with the `derive` feature in `aragors` it's little use not using it in `deadpool-arangodb` as well.

I also rebased the code on `master`

Only two issues remain: https://github.com/bikeshedder/deadpool/actions/runs/1416599665

btw. I would recommend against switching implementations via a feature gates in `aragors`. I'd rather make this explicit via `trait` or `enum` (e.g. `deadpool::Runtime`). That way you can even write an application that uses both implementations or mix crates that might depend on `aragors` with different feature flags.

Weasy666:
Yeah, i also not exactly a fan of misusing feature gates like that, but `arangors` uses them [non-additive](https://github.com/fMeow/arangors/blob/main/src/connection/mod.rs#L87-L91), so it's not possible to use multiple "conflicting" clients. So...i don't know where a trait can help us here. What exactly do you have in mind?

bikeshedder:
It's been quite some time since I last visited this. This was basically impossible to merge into the existing repository due to the workspace and the way tests were excuted. Since the removal of the workspace (#429) and a complete overhaul of the CI configuration adding this to the codebase should be a non-issue.

If anyone wants to work on this and refactor the code so it can be merged into `main` cleanly I'm happy to accept an updated PR for this.

Issue #154, opened 2021-10-13
sslmode is not respected in Config


kitsuneninetails:
Even if I set sslmode = Disable in the deadpool Config, it is not respected when the tokio_postgres Config is created in `get_pg_config`.  the reason is simple, there is no code to configure the new tokio_postgres::Config's ssl mode based on self's ssl_mode.  There are checks for all the other parameters.

bikeshedder:
Whops. I did not mean to close this issue. I'm not used to the app. 🙈

PR #155, opened 2021-10-13
Also set ssl_mode based on deadpool Config

kitsuneninetails:
Add lines to check and set ssl_mode as well.

Fixes: #154


kitsuneninetails:
Ah, I need to make SslMode public.  I don't know why it was made private along with all of `mod config` as this makes it impossible to use.  If there was a reason, please let me know, because it might invalidate my change.

PR #159, opened 2021-10-21
Fix/sync wrapper

bikeshedder:


r2d2/README.md
```diff
@@ -20,39 +20,40 @@ manager for [`r2d2`](https://crates.io/crates/r2d2) managers.
...
-async fn main() {
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let pool = create_pool(2);
     let client = pool.get().await.unwrap();
-    let answer: i32 = client
-        .interact(|client| {
+    let answer = client
+        .interact::<_, Result<i32, PgError>>(|client| {
             let row = client.query_one("SELECT 42", &[])?;
             Ok(row.get(0))
```
kaj, line 52: I think this could be written as:

```rust
        .interact(|client| client.query_one("SELECT 42", &[]).map(|row| row.get(0)))
```

... to get rid the type annotation.  But maybe it is good to show the annotation in the readme anyway, to illustrate how to write it when it is needed.

PR #161, opened 2021-11-01
add the deadpool and deadpool_sync for postgres and mysql

urkle:
When integrating deadpool-diesel into a project the examples use sqlite and pull in the Runtime from the deadpool_diesel::sqlite module.    However as I am using postgres in my app the deadpool_diesel::postgres module does not re-export those.

so this PR re-exports those for the mysql and postgres modules.  

the alternative (breaking change) would be to not export them from the DB specific modules and only re-export them from deadpool_diesel (which is being done already).

Let me know which you'd prefer and I can update this PR if needed.

### DIFF
--- a/diesel/CHANGELOG.md
+++ b/diesel/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## v0.3.1
+* Expose the reexports from deadpool and deadpool_sync for mysql and postgres
+
 ## v0.3.0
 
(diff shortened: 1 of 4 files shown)


bikeshedder:
That was indeed an oversight. Thanks for the PR.

bikeshedder:
I just published `deadpool-diesel 0.3.1` including this fix: https://crates.io/crates/deadpool-diesel/0.3.1

PR #173, opened 2021-12-30
Fix malformed ping recycle command

NathanFlurry:
Fixed regression from 0afe718d76d4cfa5d0d62e7dc7885e769baf6c63.

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -151,7 +151,8 @@ impl managed::Manager for Manager {
 
     async fn recycle(&self, conn: &mut RedisConnection) -> RecycleResult {
         let ping_number = self.ping_number.fetch_add(1, Ordering::Relaxed).to_string();
-        let n = redis::cmd(&format!("PING {}", ping_number))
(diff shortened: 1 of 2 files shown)


redis/src/lib.rs
```diff
@@ -151,7 +151,8 @@ impl managed::Manager for Manager {
 
     async fn recycle(&self, conn: &mut RedisConnection) -> RecycleResult {
         let ping_number = self.ping_number.fetch_add(1, Ordering::Relaxed).to_string();
-        let n = redis::cmd(&format!("PING {}", ping_number))
+        let n = redis::cmd("PING")
```
bikeshedder, line 154: Oh my... thanks for the fix. I messed that up. :facepalm:

bikeshedder:
I just released deadpool-redis 0.10.2 which includes this fix: https://crates.io/crates/deadpool-redis/0.10.2

This is yet another reason for implementing #174 as soon as possible. This went by unnoticed way too long.

PR #181, opened 2022-02-18
upgrade to lapin2

conradludgate:
With the recent release of lapin2, this PR updates deadpool-lapin to have support

### DIFF
--- a/lapin/src/config.rs
+++ b/lapin/src/config.rs
@@ -1,10 +1,5 @@
 use std::convert::Infallible;
 
-#[cfg(feature = "rt_async-std_1")]
-use async_amqp::LapinAsyncStdExt as _;
-#[cfg(feature = "rt_tokio_1")]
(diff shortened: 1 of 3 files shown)


lapin/src/config.rs
```diff
@@ -51,6 +46,29 @@ pub struct Config {
...
+            .field("url", &self.url)
+            .field("pool", &self.pool)
+            .field(
+                "connection_properties",
+                &ConnProps(&self.connection_properties),
+            )
+            .finish()
+    }
+}
+
```
conradludgate, line 71: `lapin::ConnectionProperties` no longer provides a Debug impl, but the fields are public, so I've implemented a wrapper for the fields that can be debugged so we don't lose any functionality

bikeshedder:
This PR is highly appreciated. Thanks a lot!

I just released deadpool-lapin 0.10.0 on crates.io: https://crates.io/crates/deadpool-lapin/0.10.0

PR #184, opened 2022-03-19
config crate upgrade

### LINKED ISSUE #183: Updating the config part in the documentation and the postgresql code example
The Config crate have updated their api and the new versions of the crate use slightly different design (the builder pattern) and names, than the one mentioned in the deadpool postgres documentation and example. 
here is the new way of setting the config:
```rust
pub struct Config {
     pub pg: deadpool_postgres::Config,
 }
 impl Config {
     pub fn from_env() -> Result<Self, config::ConfigError> {
         ::config::Config::builder()
             .add_source(::config::Environment::new().separator("__"))
             .build()?
             .try_deserialize()
     }
```
I will be happy to send a pull request with updates if required.

YounessBird:
fixes #183

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ tokio = { version = "1.0", features = ["sync"] }
 
 [dev-dependencies]
 async-std = { version = "1.0", features = ["attributes"] }
-config = { version = "0.11", default-features = false }
+config = { version = "0.12", features = ["json"] }
 criterion = "0.3"
 tokio = { version = "1.5.0", features = ["macros", "rt", "rt-multi-thread", "time"] }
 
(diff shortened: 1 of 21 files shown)


YounessBird:
This file [tests/managed_config.rs](https://github.com/bikeshedder/deadpool/pull/184/commits/c4b764dcd8824ac1f4ecc6ac51c27687ece5f0b9#diff-c7235a5f7d6e25099a1e37c9f0502e8fa73ee2a32e6d82767ff4d10000468f6f) fails to pass, I have noticed that the error produced during the review process is the same as the one I got before making the pull request, and it's due to the [Cargo.toml](https://github.com/bikeshedder/deadpool/pull/184/commits/2b44d657fa026ab1d7386e8810a3c348e49d4bf4) shared between all the packages, specifically the `default-features = false` so I had to change it to `features = ["json"]` and that seemed to solve the problem and the tests pass in my local machine. However I am uncertain as to why it still produce the same error in the review process, do you think that could be because the features parameter has to be the same for all `cargo.toml` files in the rest of the packages?

bikeshedder:
Thanks a lot for that PR! :+1:

PR #185, opened 2022-03-22
Export `TargetSessionAttrs` and `ChannelBinding`

RAnders00:
Correct me if I'm wrong, but I believe it's not possible to set these options without the library explicitly exporting these.

I've gone with exporting them under the root crate since that's where `SslMode` was already exported. It might not be a terrible option to export them under `crate::config::*` either though.

### DIFF
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## Unreleased
+
+* Export `TargetSessionAttrs` and `ChannelBinding` enums (part of `Config`
+  struct)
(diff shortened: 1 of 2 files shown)


bikeshedder:
You're 100% correct. I'm going to merge this and create a release immediately.

bikeshedder:
Thanks for the PR. :+1: 

I just pushed deadpool-postgres 0.10.2 to crates.io: https://crates.io/crates/deadpool-postgres/0.10.2

RAnders00:
Thank you ❤️

PR #207, opened 2022-07-16
CI build without --workspace

### LINKED ISSUE #190: Don't use the --workspace flag for running CI tasks
This is a blocker for #189 to be merged.

elpiel:
Fixes #190 

@bikeshedder maybe you need to allow the CI to run to see if I've fixed it? I will revert the upgrade once I fix the CI workflow.

One drawback of this is that if a new crate is introduced to the repo, it should also be added to CI matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,7 @@ jobs:
 
   clippy:
     if: ${{ github.ref == 'refs/heads/master'
-            || startsWith(github.ref, 'refs/tags/deadpool-')
-            || !contains(github.event.head_commit.message, '[skip ci]') }}
+            || startsWith(github.ref, 'refs/tags/deadpool-') }}
     strategy:
       fail-fast: false
       matrix:
@@ -41,7 +40,7 @@ jobs:
           - redis-actix-web
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
@@ -52,37 +51,17 @@ jobs:
 
   rustfmt:
     if: ${{ github.ref == 'refs/heads/master'
-            || startsWith(github.ref, 'refs/tags/deadpool-')
-            || !contains(github.event.head_commit.message, '[skip ci]') }}
-    strategy:
-      fail-fast: false
-      matrix:
-        crate:
-          - deadpool
-          - deadpool-diesel
-          - deadpool-lapin
-          - deadpool-memcached
-          - deadpool-postgres
-          - deadpool-r2d2
-          - deadpool-redis
-          - deadpool-runtime
-          - deadpool-sqlite
-          - deadpool-sync
-          # Examples
-          - example-postgres-actix-web
-          - example-postgres-benchmark
-          - example-postgres-hyper
-          - example-redis-actix-web
+            || startsWith(github.ref, 'refs/tags/deadpool-') }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: stable
           components: rustfmt
 
-      - run: cargo fmt -p ${{ matrix.crate }} --check
+      - run: cargo fmt -all --check
 
 
 
@@ -94,8 +73,7 @@ jobs:
   check-deadpool:
     name: Check deadpool
     if: ${{ github.ref == 'refs/heads/master'
-            || startsWith(github.ref, 'refs/tags/deadpool-')
-            || !contains(github.event.head_commit.message, '[skip ci]') }}
+            || startsWith(github.ref, 'refs/tags/deadpool-') }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -241,4 +277,4 @@ jobs:
           profile: minimal
           toolchain: stable
 
-      - run: cargo doc -p ${{ matrix.crate }} --all-features
+      - run: cargo doc -p ${{ matrix.crate }} --no-deps --all-features
```
elpiel, line 280: Should improve the doc check.

.github/workflows/ci.yml
```diff
@@ -29,13 +47,31 @@ jobs:
           toolchain: stable
           components: clippy
 
-      - run: cargo clippy --workspace --all-features -- -D warnings
+      - run: cargo clippy -p ${{ matrix.crate }} --no-deps --all-features -- -D warnings
```
elpiel, line 50: This should improve check time and also not fail on dependency which has clippy issues.

sqlite/Cargo.toml
```diff
@@ -23,7 +23,7 @@ serde = ["deadpool/serde", "serde_1"]
 [dependencies]
 deadpool = { path = "../", version = "0.9.1", default-features = false, features = ["managed"] }
 deadpool-sync = { path = "../sync", version = "0.1" }
-rusqlite = "0.25"
+rusqlite = "0.28"
```
bikeshedder, line 26: Please remove this change from the PR.

I'll merge https://github.com/bikeshedder/deadpool/tree/feature/rusqlite-027 separately (so it becomes part of the main branch) and then bump the `rusqlite` version to 0.28 with a separate commit. :+1:

elpiel: I will revert the upgrade once I know I've fixed the CI workflow.

bikeshedder: Gotcha. Thanks a lot for working on this. :smiley:

elpiel: Could you trigger the CI build, please?
I think because I'm an outside contributor it does not trigger the workflow and you need to allow it.

bikeshedder: For some reason I don't see any option for running the CI pipeline. Other PRs have shown this option in the past. In this case I don't get an option to do so.

bikeshedder (approved):

sqlite/Cargo.toml
bikeshedder: I tried adding a manual approval, but it still doesn't show the dialog asking if the CI pipeline should be run.

bikeshedder: Maybe it's related to this PR making changes to the CI configuration? I can see that the pipeline was run on your own clone of the repository: https://github.com/elpiel/deadpool/actions

bikeshedder: Oh... that might be the reason?

image

elpiel: totally my bad :sweat_smile:, I did a force push. Let's see how it goes.

elpiel:
Now the CI is waiting your approval

elpiel:
This should do it. I didn't realize that the example crates all start with `example-...` :smile:

.github/workflows/ci.yml
```diff
@@ -29,12 +48,31 @@ jobs:
...
 
   rustfmt:
     if: ${{ github.ref == 'refs/heads/master'
             || startsWith(github.ref, 'refs/tags/deadpool-')
             || !contains(github.event.head_commit.message, '[skip ci]') }}
```
tyranron, line 56: @elpiel additionally, would love to see if you strip this condition from CI pipeline, since GitHub now supports `[skip ci]` out-of-the-box, so they're just redundant.

.github/workflows/ci.yml
```diff
@@ -44,7 +82,7 @@ jobs:
           toolchain: stable
           components: rustfmt
 
-      - run: cargo fmt --all -- --check
+      - run: cargo fmt -p ${{ matrix.crate }} --check
```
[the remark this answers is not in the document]
elpiel: Good point! I wish there was a way for clippy too. Does it resolve dependencies? I don't like the approach of listing each of the crates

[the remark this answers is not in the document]
tyranron: @elpiel well, the way for the Clippy is `--workspace`, which doesn't work good for us, as we see. But usually it works just OK, that's why it has been done that way from the start.

bikeshedder:
> @bikeshedder I've rebased @tyranron I've fixed the PR comments you had raised_hands

For some reason I can't start the CI Pipeline. :thinking:

.github/workflows/ci.yml
```diff
@@ -18,33 +18,50 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: stable
           components: rustfmt
 
-      - run: cargo fmt --all -- --check
+      - run: cargo fmt -all --check
```
bikeshedder, line 64: I think this is a typo. It should be `--all`, no?

bikeshedder:
LGTM. Apart from one typo (`-all` -> `--all`) it looks like this should do the trick.

Could you try running the CI pipeline in your own repository? I really wonder why I can't run it here.

bikeshedder:
I can run them now, but they just report being skipped. I don't see why... As everything looks good to me I'll just merge it as is and see if it's just an issue of this PR.

bikeshedder:
After merging the branch it did execute properly. It did fail however with `clippy` and the examples. I've just commented them out for the time being: 6899e476d59c98c43f1b3f805ba74b90f4fe93ee

I'm tracking this as #208

PR #212, opened 2022-07-18
Fix `async-std` link in README

mzeitlin11:


bikeshedder:
Good catch. Thank you for the PR! :+1:

Issue #150, opened 2021-09-29
Replace HookVec::apply and apply_timeout by macros


bikeshedder:
Async functions don't support inlining:
https://rust-lang.github.io/async-fundamentals-initiative/evaluation/design/inline_async_fn.html

By using a macro for `HookVec::apply` and `apply_timeout` the latency could be decreased as the code can skip the creation and polling of additional Futures.

bikeshedder:
My benchmarks have shown that the effect of this is almost zero. I decided to stick with normal async functions for the time being.

Maybe a task for the 1.0 milestone?

bikeshedder:
After some experiments I couldn't find any performance difference. The Rust compiler already seams to properly `#[inline]` this code and/or it has such a little impact that it's just not worth it.

PR #216, opened 2022-08-21
mild linting

conradludgate:
Random source code reading, I noticed this box left around from when it used to be `Box<dyn Manager>`, no one thought to remove it. It seems harmless.

I've also broken out what clippy thought was a long function, and I seem to agree - It's slightly more self-documenting now.

Benchmarks are extremely flakey, but I notice no major improvements or regressions from this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ tokio = { version = "1.0", features = ["sync"] }
 [dev-dependencies]
 async-std = { version = "1.0", features = ["attributes"] }
 config = { version = "0.13", features = ["json"] }
-criterion = { version = "0.3", features = ["html_reports"] }
+criterion = { version = "0.3.4", features = ["html_reports", "async_tokio"] }
 itertools = "0.10.3"
 tokio = { version = "1.5.0", features = ["macros", "rt", "rt-multi-thread", "time"] }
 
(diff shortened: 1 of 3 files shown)


bikeshedder:
> Random source code reading, I noticed this box left around from when it used to be Box<dyn Manager>, no one thought to remove it. It seems harmless.

Good catch. That `Box<dyn Manager>` was necessary up until `deadpool 0.7` as it was only generic over `T` and not `M`:

- https://docs.rs/deadpool/0.7.0/deadpool/managed/struct.Pool.html

> I've also broken out what clippy thought was a long function, and I seem to agree - It's slightly more self-documenting now.

I've done that refactoring myself several times now. It always got lost in discarded branches and I never merged it into main separately. It just makes sense to finally have this changi in the main branch, too. :+1: 

> Benchmarks are extremely flakey, but I notice no major improvements or regressions from this.

It should be fine. I'm not 100% sure if that `#[inline]` hint actually does anything for `async` functions but in my own micro benchmarks I couldn't see a benefit at all. See:

- https://github.com/bikeshedder/deadpool/issues/150

conradludgate:
> It should be fine. I'm not 100% sure if that #[inline] hint actually does anything for async functions

Yeah, I didn't notice any benefit either. It would just inline the `async {}` block, it still needs to go through the generator fun.

bikeshedder (approved):

bikeshedder:
Thanks a lot! :+1:

PR #227, opened 2022-10-26
Make `Transaction`'s `statement_cache` pub

aarashy:
In the presence of changing schemas, I am hitting `cached plan must not change result type` errors in in extremely rare cases. I would like to be able to invalidate the cache and try again with just the reference to the transaction, where I don't have a handle to the `Pool` which handed out the `Transaction`. 

`statement_cache` is already pub for `ClientWrapper` and `Manager`.

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -440,7 +440,7 @@ pub struct Transaction<'a> {
     txn: PgTransaction<'a>,
 
     /// [`StatementCache`] of this [`Transaction`].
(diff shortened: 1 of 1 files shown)


bikeshedder:
This is indeed an oversight. Thanks! :+1:

bikeshedder:
I just released [deadpool-postgres 0.10.3 on crates.io](https://crates.io/crates/deadpool-postgres/0.10.3) including this fix.

aarashy:
Thanks so much!

PR #230, opened 2022-12-09
Fix Diesel recycling error handling

DGolubets:
I noticed that when DB connection was broken the pool didn't catch that.
This change fixes the problem.


bikeshedder:
Thanks a lot. :+1:

bikeshedder:
I just released `deadpool-diesel 0.4.1` on crates.io including this fix: https://crates.io/crates/deadpool-diesel/0.4.1

Issue #78, opened 2020-12-26
Improve StatementCache management methods


bikeshedder:
Right now it is not possible to clear the `StatementCache` for all clients/connections part of the pool. Thus rendering the methods `clear` and `remove` very much useless. This was recently discussed in #76 and it would be nice if there was a better way to manage the statement cache.

The easiest way to implement this would be to have a global `Version` counter as part of the `Manager`. Whenever a `StatementCache` is cleared this counter is incremented by one. Upon retrieval of statements from the statement cache this counter is checked and if a mismatch is detected the cache is cleared. This would render the recently added `remove` method unuseable though as there would be no way to invalidate just one statement unless using a `RWLock<HashSet<StatementCacheKey, Version>>` instead of a global version counter.

Another way would be to keep references to the `ClientWrapper` and/or `StatementCache` objects. As of now objects returned from the pool only know the pool they are coming from and return themselves when they are dropped. This would require all objects to be `Sync + Send` and not only `Send`.

PR #234, opened 2023-01-17
Improve statement cache perf by using `rustc-hash` instead of the `std` hasher.

LouisGariepy:
This is just a POC PR to show how the implementation could be done.

This only affects PostgreSQL, I couldn't find statement caches for the other drivers.

AFAIK, this change is non-breaking and doesn't have any API repercussions. `FxHashMap` is just a convenience type alias, it uses `HashMap` under the hood.

I'm not familiar enough with the internals of the crate to know if the current benchmarks take statement cache hashing into account, so it's hard to tell the actual real-use gains. If you think this would be necessary and could provide guidance, I could improve the PR with proper benchmarking.

I have benchmarked the hashing algorithms alone head-to-head, and `rustc-hash` performs 20% to 50% better than the std algorithm on plausible SQL queries. FxHash is particularly faster on smaller inputs. `rust-hash` is well-known to be faster, but it doesn't have DoS-safety like the std hasher has. This shouldn't be a problem since usually the app controls which queries are hashed.

## Why
Improve statement cache hashing performance without changing the API.

## Alternatives
Keep using the std hasher.

## Future work
The hashing algorithm is a small change with a small impact, but there are other ways to speed up the statement caches.

I noticed that there's a lot of `RwLock` and `Mutex` going on in the internals of the statement caches implementation. I reckon we could make significant gains using some sort of lock-free/concurrent datastructures. Most async projects are multithreaded (e.g. tokio distributes tasks on many cores if possible), so this should provide significant performance gains. I'd like to explore this in a future PR.

### DIFF
--- a/postgres/Cargo.toml
+++ b/postgres/Cargo.toml
@@ -20,11 +20,16 @@ rt_async-std_1 = ["deadpool/rt_async-std_1"]
 serde = ["deadpool/serde", "serde_1"]
 
 [dependencies]
-deadpool = { path = "../", version = "0.9.0", default-features = false, features = ["managed"] }
+deadpool = { path = "../", version = "0.9.0", default-features = false, features = [
+    "managed",
+] }
 log = "0.4"
-serde_1 = { package = "serde", version = "1.0", features = ["derive"], optional = true }
+serde_1 = { package = "serde", version = "1.0", features = [
+    "derive",
+], optional = true }
 tokio = { version = "1", features = ["rt"] }
 tokio-postgres = "0.7.2"
+rustc-hash = "1.1.0"
 
 [dev-dependencies]
 config = { version = "0.13", features = ["json"] }
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -25,7 +25,6 @@ mod generic_client;
 
 use std::{
     borrow::Cow,
-    collections::HashMap,
     fmt,
     ops::{Deref, DerefMut},
     sync::{
@@ -35,6 +34,7 @@ use std::{
 };
 
 use deadpool::{async_trait, managed};
+use rustc_hash::FxHashMap;
 use tokio::spawn;
 use tokio_postgres::{
     tls::MakeTlsConnect, tls::TlsConnect, types::Type, Client as PgClient, Config as PgConfig,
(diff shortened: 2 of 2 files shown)


bikeshedder:
> The hashing algorithm is a small change with a small impact, but there are other ways to speed up the statement caches.

Originally I wanted to create a `CachedStatement` struct which does create the hash only once (`u64`) and is `const` compatible. e.g. `const SELECT_FOO: CachedStatement = CachedStatement::new("SELECT * FROM foo");`

I'm not sure if it's actually fine to store that `u64` internally and call `hasher.write_u64(self.hash)` inside the `Hash` implementation though. As the hash is only calculated once per statement it feels wasteful to create yet another hasher and hash that `u64` just for the sake of it. Afaik that's the reason for some `IntHashMap` implementations to exist in the first place.

With that structure it would even be fine to ditch the need for calling `prepare` explicitly but just make it run on first use.  e.g.

```rust
const SELECT_FOO: CachedStatement = CachedStatement::new("SELECT * FROM foo");

async fn do_stuff(client: &Client) {
    client.query(SELECT_FOO, &[]).await.unwrap();
    ...
}
```

After some experimentation with that idea I ditched it however and went with the simpler, yet slightly less performant, way of adding `_cached` methods and using the default hashing algorithm.

There is lots of potential for bikeshedding here. :heart: :rofl: 

> I noticed that there's a lot of RwLock and Mutex going on in the internals of the statement caches implementation. 

The only reason for the statement cache to live inside a `RwLock` is that it must be possible to manage the `StatementCaches` from the via the `Manager::statement_caches` member. It's the only way to `clean` the entire cache or `remove` invividual statements from it for all connections of a given pool.

- https://github.com/bikeshedder/deadpool/issues/78

---

btw. the fastest way of working with prepared statements would be to collect all SQL that's ever going to be prepared in one large array and just use those indexes into a `Vec<Option<Statement>>`. If `Some<Statement>` is found it's prepared and ready to go and if it's `None` it needs to be prepared first.

That however would result in a VERY unergonomic API unless making heavy use of macros, parsing some local SQL files and generating the code necessary.

It's a very tempting idea but would also add an incredible amount of complexity to this crate.

bikeshedder:
I'm just coming back to this as I'm cleaning up the backlog of issues and PRs. My reasoning hasn't changed over the past two years. I'd rather change the way the API of prepared statements than making the code prone to DoS attacks.

When introducing the `CachedStatement` struct which creates the hashes at compile time this should become a non-issue anyways.

Issue #139, opened 2021-09-22
Unwind safety


dchenk:
Would it be feasible or even make sense to implement `UnwindSafe` for the `managed::Pool` type?

Some context: I've decided to replace r2d2 with deadpool-diesel for database connection pooling in an application because r2d2 has a serious flaw in the way it handles panics (see https://github.com/diesel-rs/diesel/issues/2124, https://github.com/diesel-rs/diesel/issues/2105, https://github.com/diesel-rs/diesel/pull/2020, https://github.com/sfackler/r2d2/pull/70).

The `managed::Pool` type not being `UnwindSafe` prevents certain use cases such as passing the pool to an async function that may panic but where the panic needs to be caught. As far as I can tell, because pooled objects are wrapped in a Mutex and deadpool discards poisoned mutexes, there's no risk of reusing an invalid pooled object. (At least this is the case with the `diesel` and `sqlite` implementations.)

bikeshedder:
As far as I understand the marker trait the core of deadpool should mostly be `UnwindSafe`.

I have to double check the `Drop` implementations, but if I'm not mistaken the following code should be fine to add:

```rust
impl<M: Manager, W: From<Object<M>>> UnwindSafe for Pool<M, W>
where
    M: UnwindSafe,
    W: UnwindSafe,
{
}
```

The `deadpool-diesel` and `deadpool-sqlite` crates should indeed be `UnwindSafe`. I'm confident `deadpool-postgres` and `deadpool-lapin` are `UnwindSafe`, too. Regarding `deadpool-redis` the underlying `redis` crate doesn't even support future cancellation so it's probably not `UnwindSafe` either.

dchenk:
I'm seeing a type error with the `managed::PoolInner::semaphore` field though (https://github.com/bikeshedder/deadpool/blob/master/src/managed/mod.rs#L461):
```
the type `UnsafeCell<AtomicUsize>` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
```
```
    = help: within `Pool<Manager<diesel::PgConnection>>`, the trait `RefUnwindSafe` is not implemented for `UnsafeCell<AtomicUsize>`
    = note: required because it appears within the type `tokio::loom::std::atomic_usize::AtomicUsize`
    = note: required because it appears within the type `tokio::sync::batch_semaphore::Semaphore`
    = note: required because it appears within the type `tokio::sync::semaphore::Semaphore`
[...]

PR #239, opened 2023-02-20
Implement UnwindSafe by never returning on panic

fabianfreyer:
Don't return objects to the pool on panic, since we can't be sure whether they were the source of the panic.

See-also: #139

### DIFF
--- a/src/managed/mod.rs
+++ b/src/managed/mod.rs
@@ -71,6 +71,7 @@ use std::{
     future::Future,
     marker::PhantomData,
     ops::{Deref, DerefMut},
+    panic::{RefUnwindSafe, UnwindSafe},
     sync::{
         atomic::{AtomicUsize, Ordering},
         Arc, Mutex, Weak,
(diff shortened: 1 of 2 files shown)


bikeshedder:
I think you accidentally mixed the `managed` and `unmanaged` implementations.

The `Drop` implementation of `deadpool::managed::Object` needs to be changed and the `UnwindSafe` marker needs to be added to the `deadpool::unmanaged::Pool` for both implementations to be deemed unwind safe.

fabianfreyer:
Hmm, I guess both should have the marker then?

bikeshedder:
I'm just revisiting this as part of a overall repository clean-up. Not returning objects on panic is a good idea for the managed pool. For the unmanaged pool hower this could cause some rather pesky side effects. It's unmanaged for a good reason: You create all the objects you need and then just use them forever. If a panic doesn't crash the application altogether it might end up with an empty pool waiting forever.

One way I could think about was the following:
- managed pools don't return the object if the code panicked. Though they still need to make sure to notify the pool about the slot being available!
- unmanaged pools should probably enter a `panicked` state and all future calls to `get` should panic as well. Maybe make this configurable somehow?

It's just my gut feeling that discarding objects on panic only ever makes sense if panics are handled and then you really need to make sure that the pool gets refilled. I can imagine only very few users would expect that to be the case.

PR #249, opened 2023-04-24
Add pool config option for fifo/lifo queue mode

srijs:
Implements a pool config setting that allows specifying the order in which objects are dequeued from the pool.

Besides sometimes resulting in better throughput if a "warm" object can be used, this is especially useful when wanting to automatically scale down the pool over time with decreasing usage: If the pool operates in LIFO mode, an average utilization smaller than the pool size will automatically result in some objects being recycled more often than others, allowing a sweeper process to remove the least frequently used objects.

This seems(?) to be a breaking change considering that all fields on `PoolConfig` are `pub` (see for example the `tests/managed_timeout.rs` which had to be fixed), so happy to receive guidance for how to extend the pool options without breaking anyone who might be constructing `PoolConfig` directly or exhaustively matching it. Alternatively, we could consider adding a set of methods to the pool for LIFO (e.g. `get_lifo`) but that seems somewhat less ergonomic to me.

### DIFF
--- a/src/managed/builder.rs
+++ b/src/managed/builder.rs
@@ -4,7 +4,7 @@ use crate::Runtime;
 
 use super::{
     hooks::{Hook, Hooks},
-    Manager, Object, Pool, PoolConfig, Timeouts,
+    Manager, Object, Pool, PoolConfig, QueueMode, Timeouts,
 };
 
 /// Possible errors returned when [`PoolBuilder::build()`] fails to build a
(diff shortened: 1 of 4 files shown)


bikeshedder:
Very cool feature. I'll definitely merge this for 0.10. :+1: 

Could you please rebase this PR to target the `milestone/0.10` branch?

srijs:
@bikeshedder Rebased!

bikeshedder:
Thanks a lot. :+1:

PR #251, opened 2023-05-04
Remove useless pattern matching

acatton:
This patten matching was used to unpack some struct, this can be done as the function parameter.

### DIFF
--- a/sync/src/lib.rs
+++ b/sync/src/lib.rs
@@ -121,9 +121,7 @@ where
                 f(conn)
             })
             .await
-            .map_err(|e| match e {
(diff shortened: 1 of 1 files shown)


bikeshedder:
Nice. I didn't know this was even possible. I guess it only works because `Panic` is the only variant of the enum.

btw. That PR contains an error. You probably meant to write
```rust
.map_err(|SpawnBlockingError::Panic(p)| InteractError::Panic(p))
```

acatton:
Hey I'm sorry Micheal, I didn't react on time. Thanks for fixing the commit and merging :) .

PR #253, opened 2023-05-24
Bound GenericClient with Sync to allow extension traits

rossdylan:
I ran into a problem where I could not create an extension trait based on `GenericClient` because awaiting futures returned by the `GenericClient` methods requires that `&self` be `Sync`. Both `Transaction` and `Client` are already sync, so this just makes that requirement explicit.


bikeshedder:
Thanks!

PR #256, opened 2023-06-02
Avoid panic and demonstrate how to set values for minimal example

peter-scholtens:
As the default configuration sets all fields to None it will **panic** ("No such file or directory") in all uses cases. A set of simple suggestions and the proper dependencies can help beginners.

### DIFF
--- a/postgres/README.md
+++ b/postgres/README.md
@@ -21,8 +21,15 @@ and `tokio_postgres::Transaction`.
 the `tokio1` feature of `async-std` in order to use this crate
 with `async-std`.
 
-## Example
+## Minimal Example
 
+For a minimal example you need the following dependencies in Cargo.toml:
+```
+tokio = {version = "1", features = ["rt-multi-thread","macros"] }
+deadpool-postgres = { version = "0.10" }
+tokio-postgres = "0.7"
+```
+The required configuration can be set by modifying the fields of the configuration:
 ```rust,no_run
 use deadpool_postgres::{Config, Manager, ManagerConfig, Pool, RecyclingMethod, Runtime};
 use tokio_postgres::NoTls;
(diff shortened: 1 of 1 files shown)


bikeshedder:
The minimum example doesn't panic if you've configured your local PostgreSQL accordingly. PostgreSQL by default creates a unix domain socket at `/run/postgresql/.s.PGSQL.5432` (alternatively: `/var/run/postgresql` or `/tmp`). Those are the three paths which are tried in order:

https://github.com/bikeshedder/deadpool/blob/de5157fe390edd1c332812e0af46cd4905681c60/postgres/src/config.rs#L217-L219

For this to work you just need a database user named the same as your system user and an entry in the `pg_hba.conf` like that:

```
local   all   all   peer
```

Oh, and this only works on unix-like systems. Windows doesn't have the concept of unix domain sockets and therefore doesn't support this. :see_no_evil:

bikeshedder:
Those defaults are virtually identical to `libpq`: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

Actually even `dbname` could be made optional and default to the system user name as stated in the `libpq` documentation. For some reason I decided that I didn't consider this default useful and made `dbname` mandatory instead.

How about changing the example like that:

> The following example assumes a PostgreSQL reachable via an unix domain socket and peer auth enabled for the local user in [pg_hba.conf](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html). If you're running windows you probably want to specify the `host`, `user` and `password` in the connection config or use an alternative [authentication method](https://www.postgresql.org/docs/current/auth-methods.html).

peter-scholtens:
Okay. And mention the Cargo.toml dependencies, as that was not clear to me at first too.

bikeshedder:
That should be pretty clear for anyone using Rust and Cargo, no? If you don't need any special `tokio-postgres` features you don't even need to add the `tokio-postgres` dependency.

Btw. have you seen the `postgres-hyper` example? https://github.com/bikeshedder/deadpool/tree/master/examples/postgres-hyper

I should create a new example with `axum` though as using `hyper` directly is very basic.

PR #259, opened 2023-06-28
Send UNWATCH when recycling redis connection

### LINKED ISSUE #258: Recycling Redis connection should send UNWATCH
Right now, recycling Redis connection sends `PING` with a number, and checks response.
https://github.com/bikeshedder/deadpool/blob/de5157fe390edd1c332812e0af46cd4905681c60/redis/src/lib.rs#L155-L158

This should catch usual network or server related problems. Also this would catch recycling connection after sending `MULTI` but before `EXEC`/`DISCARD`: `PING` command would return `QUEUED` instead of `PONG`. Recycling could send `DISCARD` in this case, but dropping is safe as well. `DISCARD` without `MULTI` is a error, so either state tracking or error matching would be necessary.

If connection is recycled after `WATCH` and before one of `UNWATCH`/`EXEC`/`DISCARD` key will stay watched in connection, and then next transaction in same connection could be aborted by changes in unrelated key.

IMO, Redis transactions are pretty common, and WATCH-MULTI-EXEC flow is something "usual", so it would be nice if recycling could safely handle drops anywhere during this flow, i.e. early return from `try!` between `WATCH` and `MULTI`.

`redis` crate itself is doing it in `transaction` helper. But it is for sync connections, and use-case specific.
https://github.com/redis-rs/redis-rs/blob/f0ec381bdce48c4364729228e0874ddd4e7fb5d7/redis/src/connection.rs#L1314-L1329

See https://redis.io/docs/manual/transactions/

I can send a PR with a fix and a reproducing testcase

mcheshkov:
This should allow safe recycling of connection after starting transaction with WATCH, but before EXEC/DISCARD/UNWATCH This closes #258

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -152,9 +152,13 @@ impl managed::Manager for Manager {
 
     async fn recycle(&self, conn: &mut RedisConnection, _: &Metrics) -> RecycleResult {
         let ping_number = self.ping_number.fetch_add(1, Ordering::Relaxed).to_string();
-        let n = redis::cmd("PING")
+        // Using pipeline to avoid roundtrip for UNWATCH
+        let (n,) = redis::Pipeline::with_capacity(2)
+            .cmd("UNWATCH")
+            .ignore()
+            .cmd("PING")
             .arg(&ping_number)
-            .query_async::<_, String>(conn)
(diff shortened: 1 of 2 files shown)


bikeshedder:
Thanks!

PR #263, opened 2023-07-16
Update Diesel to 2.1.0

justy777:
Updates the version of `diesel` to 2.1.0

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -26,8 +26,8 @@ serde = ["deadpool/serde"]
 [dependencies]
 deadpool = { path = "../", version = "0.9.1", default-features = false, features = ["managed"] }
 deadpool-sync = { path = "../sync", version = "0.1.0" }
-diesel = { version = "2.0.0", default-features = false }
+diesel = { version = "2.1.0", default-features = false }
 
 [dev-dependencies]
-diesel = { version = "2.0.0", default-features = false, features = ["sqlite"] }
+diesel = { version = "2.1.0", default-features = false, features = ["sqlite"] }
 tokio = { version = "1.0", features = ["macros", "rt", "rt-multi-thread"] }


weiznich:
This is not needed as the current deadpool version is by default compatible with new non-breaking diesel releases. 2.1 was just a minor release without breaking changes.

justy777:
I agree minor changes should not have breaking changes, but for Diesel 2.1.0 that is not the case.
The definition for `diesel::connection::LoadConnection` changed because they removed `ConnectionGatWorkaround`. 

So when you try to use the connection you get errors saying:
```
error[E0277]: the trait bound `&mut diesel::PgConnection: diesel::connection::LoadConnection` is not satisfied
```

weiznich:
> The definition for `diesel::connection::LoadConnection` changed because they removed `ConnectionGatWorkaround`

That's true but that's not a breaking change, sorely for the reason that `ConnectionGatWorkaround` was never part of the public API. If your code relied on this trait you used an item which was marked explicitly as not being part of the public API.


> So when you try to use the connection you get errors saying:
>
> ```
> error[E0277]: the trait bound `&mut diesel::PgConnection: diesel::connection::LoadConnection` is not satisfied
> ```

It's impossible to say something about an out of context error like that one. I would guess that you tried to pass a `&mut Connection` to a function that expects an owned connection (or anything that adds an uneven number of mutable references to either sides). I as diesel maintainer would assume an issue in your code here.

In fact you can easily see that this update is not required on the deadpool side as it sorely changes the `Cargo.toml` and the `Cargo.lock` file.

justy777:
You're right, a mutable reference gets supplied as input to the function when you call `interact` on a `SyncWrapper`. So what I get is a mutable reference of a mutable reference  when calling `.load(&mut conn)`, which does not satisfy the trait bound and results in that error message.

```rust
conn.interact(|conn| {
        let movies: Vec<Movie> = Movie::all().load(&mut conn)?;
    });
```

Thanks @weiznich for your time, I promise to take this as a learning experience.

PR #265, opened 2023-07-20
Update license field following SPDX 2.1 license expression standard

frisoft:
The new recommendation is to follow the SPDX 2.1 standard. This allows automatic license verification software to work properly. Reference: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ resolver = "2"
 authors = ["Michael P. Jung <michael.jung@terreon.de>"]
 description = "Dead simple async pool"
 keywords = ["async", "database", "pool"]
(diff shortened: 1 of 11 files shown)


frisoft:
Not sure why a simple change in the format of the license field is causing the tests to fail.
It seems there is already a problem with it.

bikeshedder:
That's caused by the latest rust and clippy version which do now recommend using `default()` instead of `Default::default()`. Clippy doesn't respect the MSRP so the only way to get rid of this is by muting those warnings. I'll take care of that in another commit.

Thanks for the PR.

PR #268, opened 2023-07-26
Allow to customize the recycle check function used in deadpool_diesel

weiznich:
This PR adds an option to customize the function used to check whether a certain diesel connection can be recycled or not. This is required to use the connection pool implementation with diesel database connections that do not support the `SELECT 1` ping query used by default. An example of such a database backend is Oracles DBMS, which requires to write `SELECT 1 FROM DUAL`. There is an existing diesel connection implementation for Oracle provided by the `diesel_oci` crate.

In addition the default recycle function was slightly changed to check for open transactions (by asking the `TransactionManager` whether it's broken) before running the ping query. This allows to discard connections with open transactions.

### DIFF
--- a/diesel/CHANGELOG.md
+++ b/diesel/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Update `deadpool` dependency to version `0.10`
 * Add `tracing` feature
+* Check for open transactions before recycling connections
+* Allow to configure a custom recycle check function to customize ping queries for different database backends
 
 ## v0.4.1
 
--- a/diesel/src/error.rs
+++ b/diesel/src/error.rs
@@ -12,6 +12,11 @@ pub enum Error {
 
     /// Failed to ping the database.
     Ping(diesel::result::Error),
+
+    /// The transaction manager of a given
+    /// connection is in a broken state. That usually
+    /// means that it contains an open uncommited transaction
+    BrokenTransactionManger,
 }
 
 impl fmt::Display for Error {
@@ -19,6 +24,7 @@ impl fmt::Display for Error {
         match self {
             Self::Connection(e) => write!(f, "Failed to establish connection: {}", e),
             Self::Ping(e) => write!(f, "Failed to ping database: {}", e),
+            Self::BrokenTransactionManger => write!(f, "Broken transaction manager"),
         }
     }
 }
@@ -28,6 +34,7 @@ impl std::error::Error for Error {
         match self {
             Self::Connection(e) => Some(e),
             Self::Ping(e) => Some(e),
+            Self::BrokenTransactionManger => None,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


weiznich:
@bikeshedder This is a kindly ping to bump up that PR on your list.

weiznich:
The CI errors look unrelated, although I likely can fix the clippy errors if that's wanted.

bikeshedder:
The only think I'm currently considering is keeping symmetry to the way this is implemented in `deadpool-postgres`. There it is an `enum` which allows specifying various premade checks and providing a custom query. It does not have an enum variant to provide your own function though.

Do you really need to provide your own function or is it just a query that needs to be issued?

Don't get me wrong. I like the idea of supporting a function. I just want to keep it in symmetry with the other deadpool crates.

bikeshedder:
How about making it an enum with the following variants:

```rs
enum RecycleMethod<C: diesel::Connection> {
    // Maybe call this Verified instead?
    // Is there a way to provide a distinction between `Fast` and `Verified`, too?
    Ping,
    CustomQuery(QueryFragment<C::Backend>),
    CustomFunction(fn(&mut C) -> Result<(), Error>),
}
```

For `deadpool-postgres` we should probably rename `Custom` to `CustomQuery` and add that `CustomFunction` in a future release, too.

bikeshedder:
Now that I think about it a `CustomQuery(String)` could be useful, too. Especially when adding config support.

As of now the `deadpool-diesel` crate doesn't support serde and has no config structs but I'd really like to add that in some future release as it just makes sense and simplifies app setup a lot.

weiznich:
I think it's reasonable do just have `CustomQuery(String)` as variant and leave the more complex variants to the `CustomFunction` version. Also it should be possible to implement a `Fast` version too. 

I will update the PR next week so that it uses such an enum instead to configure the check behavior.

weiznich:
I've implemented the suggested structure in https://github.com/bikeshedder/deadpool/pull/268/commits/52c9a14a2e8ed0b06d27d74425615857f9585658

bikeshedder:
I'm currently preparing the release of `deadpool 0.10` and all the `deadpool-*` crates with that added dependency. Could you please update your PR and maybe include this your PR, too: https://github.com/bikeshedder/deadpool/pull/259

bikeshedder:
I merged PR #259 in order to give credits. It'd be awesome if you could rebase your branch accordingly! :grinning:

weiznich:
I've rebased the PR and resolved all conflicts.

bikeshedder:
Oh my. It was late yesterday. #259 is about redis and this PR is about diesel. :facepalm: 

Merged. :+1:

bikeshedder:
Thanks a lot.

PR #271, opened 2023-09-09
Bump up MSRV to 1.63 for all crates and fix CI failures

### LINKED ISSUE #270: Fix MSRV and Clippy checks in CI pipeline
The CI pipeline currently fails because clippy doesn't respect the `rust-version` in the `Cargo.toml` and the MSRV checks fail for some other strange reasons.

tyranron:
Resolves #270 

## Background

See #270 for details.

> The CI pipeline currently fails because clippy doesn't respect the `rust-version` in the `Cargo.toml` and the MSRV checks fail for some other strange reasons.

> I think it is safe to increase the MSRV to 1.56 or even 1.59 for all crates.
> 
> Just as a reference: The current `tokio` release specifies `1.63` as MSRV.

## Solution

Bump up MSRV to 1.63, since there is no meaningful reason to keep it below `tokio`.

## Additionally

- Remove `retain_mut` polyfill usage in `deadpool` crate.
- Extend CI matrix to test and check docs for `deadpool-runtime` and `deadpool-sync` crate separately.

## Checklist

- [x] `README.md` badges updated
- [x] `CHANGELOG.md` entries added
- [x] `rust-version` is specified in `Cargo.toml`
- [x] CI pipeline updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         crate:
+          - deadpool-runtime
+          - deadpool-sync
           - deadpool
           - deadpool-diesel
           - deadpool-lapin
@@ -29,9 +31,7 @@ jobs:
           - deadpool-r2d2
           - deadpool-redis
           - deadpool-redis-cluster
-          - deadpool-runtime
           - deadpool-sqlite
-          - deadpool-sync
           # Examples
           - example-postgres-actix-web
           - example-postgres-benchmark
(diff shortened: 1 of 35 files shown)


tyranron:
@bikeshedder now it's OK.

I've added some temporary workarounds for `deadpool-lapin` and `deadpool-diesel` crates, with comments when and how to remove them in the future.

Feel free to ping me again, if new troubles with MSRV and `minimal-versions` will appear in the future.

Also, I've seen that workspace is still using 2018 edition mostly. I can make another PR with migration to 2021 edition, if you don't mind.

bikeshedder:
Awesome!

I was just wondering if we should keep the MSRV as low as possible like `1.56` or `1.59`. Both of which support edition 2021. I mean, most of the code should work with Rust `1.56` just fine. `deadpool-postgres` even works with `1.54`.

tbh. just upgrading to `1.63` eliviates a lot of pain and if someone really needs to use an older Rust version it should "just work". It just isn't officially supported anymore. :thinking: 

> Also, I've seen that workspace is still using 2018 edition mostly. I can make another PR with migration to 2021 edition, if you don't mind.

Sure. Please go ahead! :+1:

PR #286, opened 2023-11-27
Redis v0.23.4

ginkcode:
I made some changes to satisfy the connection from `redis` temporarily.
I believe they modify TcpTls to support Mutual TLS and that may need more effort to configure all kinds of TLS (native + rustls).
It's pretty awkward because this should be warned as incompatible update.


bikeshedder:
Gotcha! :+1:

extrawurst:
@ginkcode 0.24 is released. lets have a PR for that one :)

Issue #284, opened 2023-11-21
deadpool-redis-cluster: remove redis_cluster_async dependency?


brocaar:
Looking at the `redis_cluster_async` README (https://crates.io/crates/redis_cluster_async), it states:

> The functionality of this crate has been merged into the main [redis-rs](https://github.com/redis-rs/redis-rs) project.

So I think this dependency can be removed in favor of using this from the `redis` crate directly :-)

bikeshedder:
I wonder if we should keep the `deadpool-redis` and `deadpool-redis-cluster` crates separate or deprecate the `-cluster` one and add those features to `deadpool-redis` instead?

brocaar:
I think I would be in favor of a combined crate. In the end, it is targeting the same `redis` crate. This saves you probably as well the overhead of releasing this as separate crates.

Do you think it would make sense to add support for Redis Sentinel as well (https://docs.rs/redis/latest/redis/sentinel/index.html)?

shoichi1023:
I agree with including redis-cluster support in the same package as deadpool-redis crate.
I also have high expectations for this work.

bikeshedder:
This was fixed in #287 and I'm going to release a new version of `deadpool-redis` shortly.

bikeshedder:
I just release `deadpool-redis 0.14.0` to crates.io: https://crates.io/crates/deadpool-redis/0.14.0
The `deadpool-redis-cluster 0.1.1` release is marked as deprecated: https://crates.io/crates/deadpool-redis-cluster/0.1.1

PR #287, opened 2023-11-28
Merge deadpool-redis-cluster into deadpool-redis.

brocaar:
@bikeshedder this is my attempt to merge the `deadpool-redis-cluster` into the `deadpool-redis` crate + remove the `redis_cluster_async` dependency (#284). Feel free to push modifications to this pull-request :)


brocaar:
From https://github.com/bikeshedder/deadpool/pull/288:

> I do wonder if I should release a deadpool-redis-cluster 0.1.1 version that just states the deprecation or a version 0.2 which doesn't contain any code at all. The former is just a tiny nudge while the latter is a bit "aggresive" and will let anyone using the crate know about that when trying to upgrade to version 0.2.

Maybe a 0.1.1 with deprecation warnings would be best. That way the warnings are automatically printed by the Rust compiler and in the generated docs.

PR #288, opened 2023-12-07
Update to redis v0.24

aumetra:
Updates the version of `redis` used by `deadpool-redis` to v0.24

### DIFF
--- a/redis/src/config.rs
+++ b/redis/src/config.rs
@@ -124,7 +124,8 @@ impl Default for Config {
     }
 }
 
-/// This is a 1:1 copy of the [`redis::ConnectionAddr`] enumeration.
+/// This is a 1:1 copy of the [`redis::ConnectionAddr`] enumeration (excluding `tls_params` since it is entirely opaque to consumers).
+///
 /// This is duplicated here in order to add support for the
 /// [`serde::Deserialize`] trait which is required for the [`serde`] support.
 #[derive(Clone, Debug)]
@@ -175,6 +176,7 @@ impl From<ConnectionAddr> for redis::ConnectionAddr {
                 host,
                 port,
                 insecure,
+                tls_params: None,
             },
             ConnectionAddr::Unix(path) => Self::Unix(path),
         }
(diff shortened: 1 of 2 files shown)


bikeshedder:
LGTM. I just want to make a `deadpool-redis-cluster` release with an updated README stating that this crate is deprecated.

I do wonder if I should release a `deadpool-redis-cluster 0.1.1` version that just states the deprecation or a version `0.2` which doesn't contain any code at all. The former is just a tiny nudge while the latter is a bit "aggresive" and will let anyone using the crate know about that when trying to upgrade to version `0.2`.

brocaar:
@bikeshedder, this comment seems to be related to #287, not the update to redis 0.24 PR... I'll comment there...

bikeshedder:
> @bikeshedder, this comment seems to be related to #287, not the update to redis 0.24 PR... I'll comment there...

Whops. Yes. :facepalm:

redis/src/config.rs
```diff
@@ -124,7 +124,8 @@ impl Default for Config {
     }
 }
 
-/// This is a 1:1 copy of the [`redis::ConnectionAddr`] enumeration.
+/// This is a 1:1 copy of the [`redis::ConnectionAddr`] enumeration (excluding `tls_params` since it is entirely opaque to consumers).
```
bikeshedder, line 127: Are you sure about that? Maybe the `tls_params` should be exposed by our config struct as well? I feel a bit uneasy leaving out configuration fields. Even if those fields can't be deserialized with serde it should be possible to set the `tls_params` after deserialization in code.

aumetra: Why I believe that it's an oversight is that those fields here are exposed: <https://github.com/redis-rs/redis-rs/blob/3e6dc2ef611060a665d8e5d4c14bd8618517fac2/redis/src/lib.rs#L468-L469>

But yet the actual `TlsConnParams` is kept opaque.

aumetra: I created an issue on their issue tracker to hopefully clarify this: https://github.com/redis-rs/redis-rs/issues/1006

Issue #305, opened 2024-03-13
Update to Redis 0.25


brocaar:
Redis `0.25.x` has been released. Would it be possible to update `deadpool-redis` to this version?

It might be needed to make some adaptations (I have not had the time yet to look into this):

> use of deprecated method `redis::Client::get_async_connection`: aio::Connection is deprecated. Use client::get_multiplexed_async_connection instead

Maybe this means that for non-cluster connections, the connection-pool is no longer needed as these queries will use a single multiplexed connection?

brocaar:
I think a connection pool is still needed, because some queries will block the multiplexed connection, see: https://github.com/redis-rs/redis-rs/issues/851#issuecomment-1560447797

I have made these modifications in my fork:
https://github.com/brocaar/deadpool/commit/1697c5e198c72ac85c54613c5fc41aefaed09bee

Before creating a pull-request, I will do some local testing first.

brocaar:
Thanks for reviewing and merging the PR. I think this issue can be closed now :-)

Issue #102, opened 2021-06-15
Strange crosstalk(?} issue with deadpool redis


henrikrudstrom:
We have a service in development using deadpool_redis. It seems to run fine and do its thing, but after a couple of days of running it starts misbehaving: 
When the code perfoms a `GET` on a key (no matter if it exists or not), it returns the value `PONG` (we do one `PING` at startup), its not a transient issue, but it persists until we restart the service. 

```rust
let mut conn = self.pool.get().await.map_err(|e| Error::Pool(e))?;
        let query = format!("frontpage-ids/{}", frontpage_id);
        debug!(self.logger, "redis query: {}", query);
        let data: String = cmd("GET")  // <- this one returs PONG
            .arg(&query)
            .query_async(&mut conn)
            .await
            .map_err(|e| Error::Redis(format!("GET {}", query), e))?;

        deserialize_frontpage_teasers(data, max_items)
```

we initalize the pool with this code:
```rust
        let mut cfg = Config::default();
        let rurl = format!("redis://{}", rcfg.redis.url);
        cfg.url = Some(rurl.to_string());
        let pool = cfg
            .create_pool()
            .map_err(|e| Error::Redis("connect".to_string(), e))?;
        Ok(Self {
            pool: pool,
            logger: logger,
        })
```

I find this pretty mindboggling and its frustrating that the error takes many days to manifest, making debugging a *very* slow process. Any clues would be greatly appreciated...

using deadpool_redis version 0.6

bikeshedder:
This was reported as part of #97 and is caused by a known issue of the `redis` crate: 
- mitsuhiko/redis-rs#489
- mitsuhiko/redis-rs#325

Aborted futures can leave the redis connection in a unusable state.

I released patched versions of `deadpool-redis` for `0.6`, `0.7` and `0.8`. Those versions send a payload with the `PING` command and ensure that the `PONG` response return the same number.

Upgrading to [`deadpool-redis 0.6.2`](https://crates.io/crates/deadpool-redis/0.6.2) should fix that issue: https://github.com/bikeshedder/deadpool/commit/7136652b1660ee7632c4a5e453d0dec0bb083b97 https://github.com/bikeshedder/deadpool/commit/36ffd106b71781b9ffa4081c14ef30f2b7b9474d

[...]

PR #306, opened 2024-03-13
Update redis to 0.25.

brocaar:
This updates Redis to 0.25 + updates `aio::Connection` to `aio::MultiplexedConnection` (as `aio::Connection` has been deprecated). I think it depends on the use-case if people want to use the same `MultiplexedConnection` for multiple queries, or want to create new `MultiplexedConnection`s for each query (there are some queries that will block the `MultiplexedConnection`, in which case you might want to use multiple connections).

See also #305.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -28,7 +28,7 @@ cluster = ["redis/cluster-async"]
 deadpool = { path = "../", version = "0.10.0", default-features = false, features = [
     "managed",
 ] }
-redis = { version = "0.24", default-features = false, features = ["aio"] }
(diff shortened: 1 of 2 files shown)


bikeshedder:
This finally fixes the issues properly:

- bikeshedder/deadpool#102
- redis-rs/redis-rs#325

bikeshedder:
I plan on releasing `deadpool 0.11` soon including the updated `deadpool-redis` crate. I just added a follow up issue which can now be tackled:

- #307 

Thanks for the PR! :grinning: :+1:

PR #316, opened 2024-04-15
Expose API for providing a custom Connect impl and pass through the `runtime` feature

shadaj:
Currently, `deadpool-postgres` only offers APIs for creating pools that instantiate their underlying connections based on a `tokio_postgres::Config`. In some situations (mocking, custom networking), it is useful to provide an alternate connection mechanism. Thankfully, `deadpool-postgres` already has the `Client` trait for this so we just need to expose it publicly. As part of this, we also expose the existing `ConnectImpl` as a public `ConfigConnectImpl` for code that needs to abstract over a selection of connection strategy.

Our main use case for this is to enable use on WASM, where `tokio_postgres` is already supported but does not have the `tokio-postgres/runtime` feature enabled (which enables the APIs for creation based on a database URL). So this PR also adds a `runtime` feature that maps to the underlying `tokio-postgres` one for now (defaulting to true to preserve compatibility). It also includes a minor patch to disable the `keepalives_idle` config on WASM where it is not supported.

### DIFF
--- a/postgres/Cargo.toml
+++ b/postgres/Cargo.toml
@@ -31,6 +31,9 @@ tokio = { version = "1.29", features = ["rt"] }
 tokio-postgres = { version = "0.7.9", default-features = false }
 tracing = "0.1.37"
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2", features = ["js"] }
+
 [dev-dependencies]
 config = { version = "0.14", features = ["json"] }
 dotenvy = "0.15.0"
--- a/postgres/Cargo.toml
+++ b/postgres/Cargo.toml
@@ -14,7 +14,8 @@ readme = "README.md"
 all-features = true
 
 [features]
-default = ["rt_tokio_1"]
+default = ["runtime", "rt_tokio_1"]
+runtime = ["tokio-postgres/runtime"]
 rt_tokio_1 = ["deadpool/rt_tokio_1"]
 rt_async-std_1 = ["deadpool/rt_async-std_1"]
 serde = ["deadpool/serde", "dep:serde"]
@@ -27,7 +28,7 @@ serde = { package = "serde", version = "1.0", features = [
     "derive",
 ], optional = true }
 tokio = { version = "1.29", features = ["rt"] }
-tokio-postgres = "0.7.9"
+tokio-postgres = { version = "0.7.9", default-features = false }
 tracing = "0.1.37"
 
 [dev-dependencies]
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -37,11 +37,18 @@ use std::{
 };
 
 use deadpool::managed;
-use tokio::{spawn, task::JoinHandle};
+#[cfg(feature = "runtime")]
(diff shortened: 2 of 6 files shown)


postgres/Cargo.toml
```diff
@@ -14,7 +14,8 @@ readme = "README.md"
 all-features = true
 
 [features]
-default = ["rt_tokio_1"]
+default = ["runtime", "rt_tokio_1"]
+runtime = ["tokio-postgres/runtime"]
```
bikeshedder, line 18: How about making this a dependency of the `rt_tokio_1` and `rt_async-std_1` feature instead? That way the default feature set will just stay the same and it won't break for users of `rs_async-std_1` either.

postgres/src/lib.rs
```diff
@@ -82,6 +89,7 @@ pub struct Manager {
 }
 
 impl Manager {
+    #[cfg(feature = "runtime")]
```
bikeshedder, line 92: Do you really need to feature gate all of this? As the feature gated functions are generic over `T` it shouldn't cause any harm, does it? Is the core of the issue the `tokio::spawn` which isn't supported by the WASM target?

bikeshedder (commented):
LGTM. I'm just curious if we really need to feature gate so much of the crate. I would have thought that the feature gated functions would not cause any issues for the WASM platform as they are generic over T.

How can we add tests for this? It would be really nice to make sure future additions to this crate don't break WASM support.

postgres/src/lib.rs
shadaj: So this one (and its friends) need to be gated because they all initialize the Postgres connection _through_ the `tokio_postgres::Config`. That's what's not supported when you don't have the `runtime` feature on `tokio_postgres` (like on WASM), since it requires `tokio_postgres` to establish the TCP connection on your behalf. So in a world without `runtime`, we can only keep around the APIs where you bring your own connection.

postgres/Cargo.toml
shadaj: I like this idea! Will update the PR.

shadaj: Or actually, on second thought, this won't work unfortunately. Because for example on WASM, we do want `rt_tokio_1` if we're running on Tokio, just not the `runtime` feature.

shadaj:
For tests, let me look into adding CI logic to test at least compilation on WASM.

shadaj:
In theory, the new CI config should test compilation on WASM!

bikeshedder:
I just removed the `runtime` feature replacing it by conditional dependencies instead:

```toml
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
tokio-postgres = "0.7.9"

[target.'cfg(target_arch = "wasm32")'.dependencies]
tokio-postgres = { version = "0.7.9", default-features = false }
```

Could you please check if that works for you?

This should be a non breaking release for non-WASM targets now.

shadaj:
@bikeshedder oh actually I think this doesn't work, due to https://github.com/rust-lang/cargo/issues/1197

In our downstream crates we are seeing the default features pulled in when compiling on WASM.

Issue #338, opened 2024-06-25
Redis Sentinel connection pool


smf8:
Hello,
Is there a plan to add the sentinel client connection pool support to the `deadpool-redis` crate?

Thanks in advance

smf8:
I'm fairly new to Rust, so I apologize beforehand if I ask obvious questions.

Based on current implementations of `Client` and `ClusterClient`. I think by combining changes from cluster and normal clients we can create a `Sentinel` client.

The only issue I've faced is that [`Manager.get_async_connection()`](https://github.com/bikeshedder/deadpool/blob/master/redis/src/lib.rs#L150) requires [`&mut self`](https://github.com/redis-rs/redis-rs/blob/main/redis/src/sentinel.rs#L769).
I didn't find any workaround except using a `Send + Sync` mutex (e.g. `tokio::sync::Mutex`) Which introduces a new dependency and additional locking/unlocking for connection creation.

So it will look like this:
```rust
pub struct Manager {
    client:  Mutex<SentinelClient>,
    ping_number: AtomicUsize,
}

async fn create(&self) -> Result<MultiplexedConnection, RedisError> {
    let mut client = self.client.lock().await;
    let conn = client.get_async_connection().await?;
    Ok(conn)
}

```

Is there any better way of achieving this?

bikeshedder:
You can add a `tokio::sync::Mutex` without introducing any additional dependencies. `deadpool` itself already uses the `tokio::sync::Semaphore` internally (which is runtime agnostic). No other parts of `tokio` is a requirement though as is explained in the README:

https://github.com/bikeshedder/deadpool/blob/7b933c585d39f2ef57304353a35e139174b28f8e/README.md?plain=1#L181-L190

If you wanted to avoid the use of the `tokio::sync::Mutex` you could use a `tokio::sync::Semaphore` with a single permit and store the client inside a `std::cell::Cell`. Though I'd just go with the `Mutex` as it's just a bit cleaner and easier to reason about.

As far as I know the Rust compiler still compiles the entire `tokio::sync` module even if only the `Semaphore` is used. So using the `Mutex` from tokio shouldn't add any additional compile time.

bikeshedder:
I forgot to answer your question...

> Is there any better way of achieving this?

No. I think that's the way to go.

Feel free to create a PR adding this feature to `deadpool-redis`.

bikeshedder:
I just published `deadpool-redis 0.17.0` on crates.io including your changes:

[...]

PR #339, opened 2024-07-16
Add redis sentinel connection pool

smf8:
Hello again, I've come up with this draft to add a sentinel connection pool according to #338.

It's currently in a draft state, meaning it might have some issues but I wanted to share the general idea with you. It'd be great if you could review this and point out whether I missed anything or if anything could be improved.

I apologize again if I made any rookie mistakes but I'm fairly new to Rust and I'd like to contribute as well as I can.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,10 +2,10 @@ name: CI
 
 on:
   push:
-    branches: ["master"]
-    tags: ["deadpool-*"]
+    branches: [ "master" ]
+    tags: [ "deadpool-*" ]
   pull_request:
-    branches: ["master"]
+    branches: [ "master" ]
 
 env:
   RUST_BACKTRACE: 1
@@ -87,8 +87,8 @@ jobs:
           toolchain: stable
 
       - run: cargo check -p deadpool
-                   --no-default-features
-                   --features ${{ matrix.feature1 }},${{ matrix.feature2 }}
+          --no-default-features
+          --features ${{ matrix.feature1 }},${{ matrix.feature2 }}
 
   check-integration:
     name: Check integration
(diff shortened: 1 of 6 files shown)


bikeshedder:
LGTM except for the linter and doc errors.

btw. you also forgot to change the comment in the source code from `cluster` to `sentinel`.

It would be nice if there was a way to share the configuration from `redis`, `redis::cluster` and `redis::sentinel` somehow. One way would be to add support for multiple URLs for the regular redis client as well. It could just try the provided connection infos in order and fail if neither one worked. That way one config could be used for all three implementations. :thinking:

smf8:
Hi, Thank you for the review. Sorry, I've postponed some lint/doc issues for later and wanted to see if the general idea looked good.

I'll try to address these issues alongside the change in configuration struct this weekend and update this PR.

smf8:
Hi again,
> One way would be to add support for multiple URLs for the regular redis client as well. It could just try the provided connection infos in order and fail if neither one worked. That way one config could be used for all three implementations.

Won't this break the syntax for old users? (changing fields of regular redis `Config` struct) since the raw config formats (e.g. json, yaml, env, etc.) used for serde serialize/deserialize differ for `Option<String>` and `Option<Vec<String>>`?

bikeshedder:
> Won't this break the syntax for old users? (changing fields of regular redis `Config` struct) since the raw config formats (e.g. json, yaml, env, etc.) used for serde serialize/deserialize differ for `Option<String>` and `Option<Vec<String>>`?

True. That's why we keep a changelog and document changes like that. The change should be made in a way that either doesn't break existing configurations or breaks properly. By "breaking properly" I mean the user should be made aware that the configuration code returns an error if the old format is still being used rather than just ignoring the old values. If this is not feasible it could also be added as a deprecated option that only returns an error if both variants are being used.

This is nothing critical though. Feel free to duplicate the config structures for now and we'll clean this up at a later point. It's not a blocking issue with the code.

smf8:
Hi, I've tried fixing the CI pipeline. As for the config comment, I'll try sending another PR for that if that's OK.

Issue #350, opened 2024-08-30
Support for read_from_replicas option in deadpool_redis


fruscianteee:
Hello,

I'm currently using the redis-rs crate, which provides a read_from_replicas option. When this option is enabled, read queries will be sent to the replica nodes, while write queries will be sent to the primary nodes.

I would like to know if there is a way to enable the read_from_replicas option when using deadpool_redis. If this feature is not currently supported, are there any plans to add support for it in the future? 

Thank you for your assistance.

fruscianteee:
I believe it can be enabled with something like this:
```rust
let client = ClusterClientBuilder::new()
    .read_from_replicas()
    .build()?;
```
So, simply calling it on the builder should be sufficient.

> I'm always happy to accept PRs if you want to implement this yourself.

Thank you for the opportunity. This would be my first contribution, but I'd like to give it a try. I'll review the source code in `deadpool/redis/src/cluster` and come up with a proposal for the changes.

fruscianteee:
Thank you for reviewing the PR. It was a great experience, and I really enjoyed it. I’ll be happy to contribute again if there’s anything else in the future.

With that, I’d like to close this issue.

bikeshedder:
I just published `deadpool-redis 0.17.0` on crates.io including your changes:

- :package: https://crates.io/crates/deadpool-redis/0.17.0

Thanks a lot!

PR #351, opened 2024-09-08
Add read_from_replicas method and environment variable support for Redis cluster configuration

fruscianteee:
## Overview

This PR introduces the following changes:
1. Added the `read_from_replicas` method to the `Config` struct for easier configuration.
   - The `read_from_replicas` method allows enabling read operations from Redis replica nodes in a more intuitive way.
2. Enabled reading the `REDIS_CLUSTER__READ_FROM_REPLICAS=true` environment variable for configuration.
   - The configuration now supports environment variables such as `REDIS_CLUSTER__READ_FROM_REPLICAS` for enabling or disabling reading from replicas.

## Changes
- Implemented the `read_from_replicas` method to simplify the process of enabling replica reads.
- Updated the configuration handling to support reading environment variables, specifically `REDIS_CLUSTER__READ_FROM_REPLICAS`.

## Testing
- Added a test to ensure that Redis write operations work correctly when `read_from_replicas` is enabled.
- Skipped testing whether read commands are actually sent to replicas, as this behavior is managed internally by `redis-rs`, and we rely on its implementation for this verification.

## Issue
This PR addresses issue #350.

### DIFF
--- a/r2d2/CHANGELOG.md
+++ b/r2d2/CHANGELOG.md
@@ -9,9 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- __Breaking:__ Modified the `Config` struct to support `read_from_replicas` for Redis clusters, using the redis-rs crate's functionality.
-- Added support for enabling `read_from_replicas` via the environment variable REDIS_CLUSTER__READ_FROM_REPLICAS.
-
 ## [0.4.1] - 2024-05-04
 
 - Update `deadpool` dependency to version `0.12`
--- a/r2d2/CHANGELOG.md
+++ b/r2d2/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


bikeshedder:
Could you please look into the clippy error and remove the superfluous `#[must_use]`. Actually I would prefer for the `read_from_replicas` method to be removed. Writing `cfg.read_from_replicas = true` is just as easy as calling a `cfg.read_from_replicas()` method.

bikeshedder:
Is there a reason to keep the `read_from_replicas()` method? I'd rather remove it if it doesn't add any value. It just adds multiple ways to do the same thing.

bikeshedder:
Could you also add this to the `CHANGELOG.md` of the `deadpool-redis` crate? That'd be awesome. :partying_face: 

I was about to make a `deadpool-redis` release. As this changes a public structure it needs to be a "breaking release" even though I don't think anyone uses the cluster config directly.

fruscianteee:
OK! I will also add the changes to the CHANGELOG.md, so please wait a moment.

bikeshedder:
You changed the `r2d2/CHANGELOG.md` by accident. That should've been added to the `redis/CHANGELOG.md` instead. :see_no_evil:

fruscianteee:
Oh no! I’m sorry! I’ve fixed it now!

PR #355, opened 2024-09-19
Fix serde not in scope error.

brocaar:
As serde is an optional dependency, the #[serde(default)] must be behind a cfg_attr to make sure it is only applied in case the serde feature is enabled.


bikeshedder:
Whopsies! Thank you for the fix! :+1:

bikeshedder:
I just released `deadpool-redis 0.17.2` at crates.io which includes this fix:

- :package: https://crates.io/crates/deadpool-redis/0.17.2

PR #357, opened 2024-09-22
Allow configuring Redis connections via AsyncConnectionConfig

Louis-Tarvin:
Currently, deadpool-redis doesn't provide any way to modify the [`MultiplexedConnection`](https://docs.rs/redis/0.27.2/redis/aio/struct.MultiplexedConnection.html) configuration (e.g. to set connection or response timeouts).
This PR addresses this by implementing a new constructor for the `Manager`, `new_with_connection_config`, that lets the user provide an [`AsyncConnectionConfig`](https://docs.rs/redis/0.27.2/redis/struct.AsyncConnectionConfig.html) to the `Manager`, such that we can construct the connection via [`get_multiplexed_async_connection_with_config`](https://docs.rs/redis/0.27.2/redis/struct.Client.html#method.get_multiplexed_async_connection_with_config).

Please let me know if there's anything I have missed or if you'd rather this be done differently.

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -36,7 +36,7 @@ use std::{
 use deadpool::managed;
 use redis::{
     aio::{ConnectionLike, MultiplexedConnection},
-    Client, IntoConnectionInfo, RedisError, RedisResult,
+    AsyncConnectionConfig, Client, IntoConnectionInfo, RedisError, RedisResult,
 };
 
 pub use redis;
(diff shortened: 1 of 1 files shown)


redis/src/lib.rs
```diff
@@ -143,6 +153,23 @@ impl Manager {
...
+            connection_config: AsyncConnectionConfig::default(),
+        })
+    }
+
+    /// Creates a new [`Manager`] from the given `params` and [`AsyncConnectionConfig`].
+    ///
+    /// # Errors
+    ///
+    /// If establishing a new [`Client`] fails.
+    pub fn new_with_connection_config<T: IntoConnectionInfo>(
```
bikeshedder, line 165: I think I would like the name `from_config` or `from_async_connection_config` better for this. What do you think?

Louis-Tarvin: `from_config` seemed more consistent with how things are named elsewhere, so I've changed it to that 👍

PR #368, opened 2024-11-19
Fix deadpool-memcached

bikeshedder:
The current `deadpool-memcached` release is somewhat broken. I've just added tests to make sure it doesn't break in the future.

I'm planning to add a config structure as well so it is on-par with the other `deadpool-*` crates.


ederuiter:
Build failures seem unrelated to these changes; they all fail building a dependency
```
error: could not compile `value-bag` (lib) due to 3 previous errors
```

Issue #226, opened 2022-10-19
Redis PubSub


Owez:
There's a [`.publish()`](https://docs.rs/deadpool-redis/0.10.2/deadpool_redis/struct.Connection.html#method.publish) method in deadpool redis but I can't find any way to subscribe to channels for pubsub usage in redis. How should I go about this, and are there any async alternatives if deadpool doesn't support it?

bikeshedder:
`deadpool_redis::Connection` is just a wrapper around [`redis::aio::Connection`](https://docs.rs/redis/latest/redis/aio/struct.Connection.html) which doesn't support `into_pubsub` as it requires ownership of the object and not just a reference to it. Most users don't need shdrt lived `PubSub` objects and therefore I never really bothered implementing a solution for this. Just create the object directly and keep it for the entire duration of the program. (or use `Connection::take` as explained below)

It could be possible to provide this feature by implementing some clever wrapping utilizing the [`redis::aio::Connectioninto_pubsub`](https://docs.rs/redis/latest/redis/aio/struct.Connection.html#method.into_pubsub) and [`redis::aio::PubSub::into_connection](https://docs.rs/redis/0.21.4/redis/aio/struct.PubSub.html#method.into_connection) methods. If this is indeed a valid use case of yours feel free to submit a PR for this feature.

In the mean time the easiest way to get a `PubSub` object is by calling `Pool::get` followed by a `Object::take` thus detaching the object from the pool and taking full ownership over the underlying `redis::aio::Connection` object:

```rust
let obj = deadpool_redis::Connection::take(pool.get().await?);
let pubsub = obj.into_pubsub();
```

Due to the `take()` call the object just won't be returned to the pool afterwards and there is currently no way to manually add a connection back to the pool.

bikeshedder:
I'm closing this as I don't think it's very practical to use a pool for listening for messages. I might be totally wrong so please feel free to comment here and/or open a new issue describing a use case where this is needed.

Owez:
Yep your right, this ended up being an issue with the redis pubsub implementation itself. I think it was a third party library which was merged into it?

sp1ff:
Would [keyspace notifications](https://redis.io/docs/manual/keyspace-notifications/) be a use case? I need to know about changes to a particular key, but once they've happened, I'm done with it.

bbaldino:
[...]

PR #369, opened 2024-11-26
Allow to specify node_connection_info for redis sentinel

### LINKED ISSUE #367: Unable to switch to a different redis database when using deadpool-redis with sentinel
I tried to add support to chirpstack for redis sentinal (https://github.com/chirpstack/chirpstack/pull/553) 

However when I tried it with our redis configuration it fails with errors like:

```
Error occurred while creating a new object: Redis server refused to switch database - ResponseError: unknown command 'SELECT', with args beginning with: '1' : Redis server refused to switch database - ResponseError: unknown command 'SELECT', with args beginning with: '1' 
```

It seems that the sentinel client tries to select the database on the sentinel and not on the actual underlying redis connection.
Currently to configuration is pretty simple:

```rust
        let pool = deadpool_redis::sentinel::Config::from_urls(
            conf.redis.servers.clone(),
            conf.redis.master_name.clone(),
            deadpool_redis::sentinel::SentinelServerType::Master,
        )
        .builder()?
        .max_size(conf.redis.max_open_connections as usize)
        .build()?;
```

In this case the `conf.redis.servers` hold a single value with `redis://sentinel:26379/1`

Maybe there is another way of configuring the database for the redis connection; but this is how other clients like go-redis allow you to configure the database.

Is there any way to achieve this with the current version or is this a bug?

ederuiter:
This allows you to select a different database/username/password for the connections to the underlying redis servers.

Closes #367

### DIFF
--- a/redis/src/sentinel/config.rs
+++ b/redis/src/sentinel/config.rs
@@ -1,8 +1,6 @@
-use redis::sentinel::SentinelNodeConnectionInfo;
-use redis::TlsMode;
-
 pub use crate::config::ConfigError;
-use crate::{ConnectionAddr, ConnectionInfo};
+use crate::{ConnectionAddr, ConnectionInfo, RedisConnectionInfo};
+use serde::{Deserialize, Serialize};
 
 use super::{CreatePoolError, Pool, PoolBuilder, PoolConfig, Runtime};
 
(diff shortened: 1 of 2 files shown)


ederuiter:
Build failures seem unrelated to these changes; they all fail building a dependency
```
error: could not compile `value-bag` (lib) due to 3 previous errors
```

bikeshedder:
Yes, those errors are completely unrelated.

It's quite a bummer that the dependencies keep breaking the MSRV checks of the deadpool crates. :disappointed:

ederuiter:
Makes sense; I amended my commit with the changes

ederuiter:
@bikeshedder would it be possible to release a new version of deadpool-redis?

bikeshedder:
I'm currently hesitant releasing a version with #357 being merged which adds a feature that I'm going to rework as part of  #226. I could either revert #357 and make a release that only adds this PR or release a version that also includes a change that is going to be reverted/reworked.

I'm thinking of going the middle ground. Reworking #357 so it just adds the timeout config which can be kept even when PubSub support lands in `deadpool-redis`. :sweat_smile:

I just need a moment to make those changes. I'll release a new version of `deadpool-redis` asap.

Issue #380, opened 2025-02-02
Runtime resizing causes frequent panics


Qqwy:
Minimal reproducible example: https://github.com/Qqwy/rust-deadpool-resize-mre/blob/main/src/main.rs

The idea of this code is to increase or decrease a pool size at runtime, based on the results of earlier usage of the pool.

### Expected behaviour:

This code should run to completion, and we should see the pool grow slowly as more and more connections are added.

### Actual behaviour

After the first few connections have been added, the code panics with

```
thread 'tokio-runtime-worker' panicked at /home/marten/.cargo/registry/src/index.crates.io-6f17d22bba15001f/deadpool-0.12.1/src/managed/mod.rs:493:30:
attempt to subtract with overflow
```

after which any other tasks, when attempting to fetch a connection from the pool, also panic with a `PoisonError` (this is what it means for a mutex to be poisoned after all). 
This in turn causes Tokio to either abort or to hang the program (non-deterministically).

---

It seems that the calculation of the new pool size in `Pool::resize` is prone to a race-condition that can cause the subtraction to underflow.
(The stack trace complains about [this line](https://github.com/bikeshedder/deadpool/blob/master/src/managed/mod.rs#L493))

Qqwy:
Changing

```rust
// grow pool
if max_size > old_max_size {
    let additional = slots.max_size - slots.size;
    slots.vec.reserve_exact(additional);
    self.inner.semaphore.add_permits(additional);
}
```

to

```rust
// grow pool
if max_size > old_max_size {
    if slots.size < max_size {
        let additional = slots.max_size - slots.size;
        slots.vec.reserve_exact(additional);
        self.inner.semaphore.add_permits(additional);
    }
}
```

seems to resolve the problem.
However, it is very surprising to me that `size` can be larger than `max_size`, because clearly this piece of code relies on that assumption.

bikeshedder:
I just had to add a new label to the project: `P-critical`

Even though it's probably not affecting a lot of people just using the deadpool API should not cause any panics.

Looking at the code I see the bug. `slots` can be larger than `max_size` as objects still borrowed from the pool are still considered `slots`. Only when an object is actually dropped and returned to the pool it is seen as garbage and discarded and the number of slots reduced.

I think the fix is as simple as this:

```diff
         // grow pool
         if max_size > old_max_size {
-            let additional = slots.max_size - slots.size;
+            let additional = slots.max_size - old_max_size;
[...]

PR #382, opened 2025-02-02
Fix panics on resize

bikeshedder:
See #380


bikeshedder:
@Qqwy Some of the other resize tests are failing now. I guess the tests itself make some wrong assumptions now. 

I need to leave now. If you want to have a look any help would be highly appreciated! 👍

bikeshedder:
I've cleaned up the branch. If all tests pass (except for the MSRV tests which are currently broken) I'll merge the branch and prepare a release.

In theory this is a breaking release due to the added return type of `Pool::retain` but it's highly unlikely that anyone did something silly like `let x: () = pool.retain(...);`. It should be pushed to anyone using the current version of deadpool so I'm going ahead and include this change as well in this release.

PR #383, opened 2025-02-02
Fix panics on resize

Qqwy:
_(As I don't have write access to the Deadpool repo directly, I had to continue on #382 by creating a new PR on my fork)_

@bikeshedder I believe the cause of the broken tests was that you used `max_size` rather than `old_max_size` in the fix. Probably just a typo.

This PR:
- [x] Fixes that typo
- [x] Moves all resizing-related tests to the same testing submodule
- [x] Mentions the issue that the new test is a regression test for in the new test's comments.

### DIFF
--- a/src/managed/mod.rs
+++ b/src/managed/mod.rs
@@ -490,7 +490,7 @@ impl<M: Manager, W: From<Object<M>>> Pool<M, W> {
         }
         // grow pool
         if max_size > old_max_size {
-            let additional = slots.max_size - max_size;
+            let additional = slots.max_size - old_max_size;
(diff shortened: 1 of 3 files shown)


bikeshedder:
Oh my. Coding with time pressure is always a great idea. 😂

bikeshedder:
And that's why I always write unit tests / reproductions first and then fix the bugs. 😅

Qqwy:
Don't be too harsh on yourself. It shows that the test suite of Deadpool is of high quality :grin: 

Thank you very much for your efforts to get this identified and fixed quickly! :smiling_face_with_three_hearts:

Issue #379, opened 2025-01-28
Example of connecting to a Redis Server using TLS


spirrello:
Hello,

Could you kindly provide an example of how to use deadpool with a Redis server that has TLS enabled?  I tried to follow https://github.com/bikeshedder/deadpool/issues/311 but it wasn't clear.

Here is a sample of my code:

```
async fn setup_pool(redis_host: String, redis_username: String, redis_password: String) -> Pool {
    let config = deadpool_redis::Config {
        url: None,
        connection: Some(ConnectionInfo {
            addr: ConnectionAddr::TcpTls {
                host: redis_host,
                port: 6379,
                insecure: true,
            },
            redis: RedisConnectionInfo {
                db: 0,
                username: Some(redis_username),
                password: Some(redis_password),
                protocol: deadpool_redis::ProtocolVersion::RESP2,
            },
        }),
        pool: Some(deadpool_redis::PoolConfig {
            max_size: 16,
            timeouts: Timeouts {
                wait: Some(Duration::from_secs(5)),
                create: Some(Duration::from_secs(5)),
                recycle: Some(Duration::from_secs(5)),
            },
            ..Default::default()
        }),
    };

    config.create_pool(Some(Runtime::Tokio1)).unwrap()
}

async fn get_key(pool: Pool, key: String) {
    let mut conn = pool.get().await.unwrap();    // Errors out here
    let value: String = cmd("KEY")
        .arg(&[&key])
        .query_async(&mut conn)
        .await
        .unwrap();

    log::info!("key: {}, value: {}", key, value);
}
```

Here is the error I am getting:

```
called `Result::unwrap()` on an `Err` value: Backend(Cannot connect to TCP with TLS without the tls feature- InvalidClientConfig)
```

Here is my Cargo.toml

```
[dependencies]
tokio = { version = "1", features = ["full"] }
redis = { version = "0.28.2", features = ["tokio-native-tls-comp"] }
log = "0.4.25"
env_logger = "0.9"
anyhow = "1.0.95"
deadpool-redis = "0.18.0"
```

bikeshedder:
Could you share how you solved this? Is there anything that needs to be added to the `deadpool-redis` documentation?

spirrello:
> Could you share how you solved this? Is there anything that needs to be added to the `deadpool-redis` documentation?

[...]

Issue #169, opened 2021-11-10
Re-export features of dependencies


bikeshedder:
This has been asked before. I can't seam to find the issue right now so I'm writing this down as a reminder.

Right now when using `deadpool-postgres` and using it together with `tokio-postgres` + `with-uuid-0_8` there is no other way than adding both `deadpool-postgres` and `tokio-postgres` (+ `with-uuid-0_8`) to the `dependencies` while making sure that the versions match.

It was many times before if `deadpool-postgres` could just reexport all the feature flags of `tokio-postgres`. I have no absolute oppionion here and in the end decided against it as it's yet one more thing to maintain.

Just looking at `tokio-postgres` it is almost a no-brainer to re-export the features. Though when you look at  `deadpool-sqlite` it would require 35 feature flags if I'm counting correctly.

https://github.com/rusqlite/rusqlite/blob/master/Cargo.toml - There are 30 feature flags and 7 optional dependencies. I believe two of the optional dependencies are not feature related (`lazy_static` and `byteorder`).

It would be so much easier if Cargo just had a way of enabling features of dependencies of dependencies like that:
```toml
deadpool-postgres = { version = "0.10", features = "tokio-postgres/with-uuid-0_8" }
```

A somewhat related `cargo` issue is this: https://github.com/rust-lang/cargo/issues/9094

While this is not possible, I wonder if `deadpool-*` should work around this and re-export the features so the users of the `deadpool-*` crates don't have to import the backend crates themselves in order to activate features from them.

--- 

Another way to solve this would be to make the backend version explicit. e.g. `deadpool-postgres` in version 0.10 does have a feature called `tokio-postgres-0_7` which enables the `deadpool_postgres::tokio_postgres_0_7` module:

```toml
deadpool-postgres = { version = "0.10", features = ["tokio-postgres-0_7"] }
tokio-postgres = { version = "0.7", features = ["with-uuid-0_8"] }
```

This would make the versions 100% obvious to reason about and as a nice side-effect decouple the versions of `deadpool-postgres` from `tokio-postgres`. Right now whenever a breaking version of `tokio-postgres` is released a breaking version of `deadpool-postgres` must be released as well.

Big downside here would be that users of `deadpool-postgres` now have to use imports like  `use deadpool_postgres::tokio_postgres_0_7::Pool` as the root of `deadpool_postgres` now contains no public exports whatsoever.

[...]

PR #389, opened 2025-02-17
Expose 'redis' features

woshilapin:
Following up on #379 and the desire of re-exposing at least TLS features from `redis` in order to avoid maintaining both `deadpool-redis` and `redis` dependency (and ensuring the versions are correct).

The first commit just do that: re-exporting all the features of `redis` by prefixing them with `redis-`.
The second commit is only fixing a missing feature in the documentation (I can open another PR if you prefer to keep things separated)
The third commit is an attempt to alleviate the maintenance burden by letting the CI ensuring all the `redis` features are always re-exposed in `deadpool-redis`. It's only a proposition at this stage. It could be made more generic to support all the other crates of `deadpool` (#169), or it could be removed since the maintenance burden is now in those scripts 🤷


bikeshedder:
That CI check is awesome!

bikeshedder:
The "How can I enable features of the `redis` crate?" entry in the `FAQ` can now be removed, too.

I wonder if we really should use the prefix `redis-` for re-exported features. I want to keep the `deadpool-*` crates in line and with `tokio-postgres` it would become an unwieldy `tokio-postgres-with-...` feature name. I'm tempted to go with the `tokio-postgres` approach of prefixing features that enable dependencies `with-*` or just go with the original feature name without a prefix at all.

What are your thoughts on that? Consistency is key and it just makes sense to apply this to all `deadpool-*` crates in the future.

bikeshedder:
Features such as `cluster` and `sentinel` are a great example. It makes very little sense to enable the `redis/sentinel` feature without also enabling the `deadpool-redis/sentinel` feature. I guess not prefixing the features makes the most sense.

The only features that the `deadpool-*` define themselves are the `rt_*` ones, `serde` and `tracing`.

It's very unlikely that this is ever going to cause a naming conflict. e.g. if you enable `tracing` on a `deadpool-*` crate you probably also want to enable it for the underlying crate anyways.

My vote is on not adding a prefix at all.

woshilapin:
> The "How can I enable features of the redis crate?" entry in the FAQ can now be removed, too.

Indeed, done.

> The only features that the deadpool-* define themselves are the rt_* ones, serde and tracing.

I assumed you mean `tokio` instead of `tracing`. By the way, I made it more explicit in the list of features (instead of relying on `optional = true` that makes it a feature) because it helps for the CI check.

> My vote is on not adding a prefix at all.

I agree, let's see if we can make this work with this CI check.

> I wonder if we really should use the prefix redis- for re-exported features. I want to keep the deadpool-* crates in line and with tokio-postgres it would become an unwieldy tokio-postgres-with-... feature name. I'm tempted to go with the tokio-postgres approach of prefixing features that enable dependencies with-* or just go with the original feature name without a prefix at all.

In theory, I could go either way. But in practice, I actually used the prefix `redis-` for the CI check in [here](https://github.com/deadpool-rs/deadpool/pull/389/files#diff-cbf18fece0e14ce100ea5815bc501310a9e7637d4f6dad74ad7cf113e143859cR5-R7).

Basically, the problem is the following: you have the list of `redis` features (from which you remove `default` and the deprecated ones). You also have the list of `deadpool-redis` features (all the ones that are prefixed by `redis-`). How can you be sure that all of them are exposed in `deadpool-redis`? The approach I had in my first version is to compare the 2 lists (sorted and uniqueness). If they're different, we need the CI to tell us.

Now, if we remove the prefix, getting the list of `deadpool-redis` features means getting all the features (also `rt_tokio_1`, `rt_async-std_1`, `serde` and `tokio`). A simple `diff` won't work because `deadpool-redis` has its own specific features that are not present in `redis`. Of course, we could change "redis features compared to deadpool-redis features" to " redis features is a subset of deadpool-redis features"... but, what if `redis` introduce a `serde` feature? The CI won't yell because it would still be a subset... however, you probably would want to add `redis/serde` to the `serde` feature in `deadpool-redis`.

For this reason, I made a slightly more annoying solution where we have to maintain a list of the specific features of `deadpool-redis` in a file. I agree, this is a solution for an edge case with a very low probability of happening. We could drop that and just expect a bug report when it happens, I'm honestly fine with both solutions.

woshilapin:
⚠️ It might be a breaking change since before, we had `cluster = ["redis/cluster-async"]` but I then changed it to

```
cluster = ["redis/cluster"]
cluster-async = ["redis/cluster-async"]
```

> By the way, I made it more explicit in the list of features (instead of relying on optional = true that makes it a feature) because it helps for the CI check.

I actually made the feature `tokio` explicit now with a `tokio = ["dep:tokio"]`, thinking it was needed for the CI, but I'm not so sure anymore. Do you mind keeping it? Or do you prefer the `optional = true` only?

bikeshedder:
## `cluster` vs `cluser-async`

I'd make `cluster-async` an alias to `cluster` (or vice verse). `deadpool` is all about async pools that `-async` prefix shouldn't be necessary.

## `dep:tokio`

That optional tokio dependency is not meant to be used as a dependency. It's only there because the `sentinel` feature which depends on `tokio/sync`.

Sadly all optional dependencies are automatically exported as features. I hope in a future version of cargo this gets removed and/or there is a way to disable this behavior.

woshilapin:
OK, therefore, I removed the `tokio` features (better hide it a little if possible if I understand correctly), and I made `cluster-async` an alias to `cluster` (see https://github.com/deadpool-rs/deadpool/compare/0faac99b8a5a285832e10e07b3cf326b1543942d..2f71d20562018f9dfebe17cd117a5e0c2bb36c7c).

bikeshedder:
Every optional dependency is exposed as feature. Enabling the `tokio` feature doesn't do any harm except that it adds an additional dependency which doesn't get used unless the `sentinel` feature is also enabled. So no need to take any action here.

This is just something that's on my wishlist ever since working features and optional dependencies in rust crates.

bikeshedder:
I just learned something new.

If a feature exists that depends on an optional dependency via the `dep:` prefix that optional dependency no longer becomes an implicit feature.

`dep:tokio/sync` does not work but Cargo gives a great error message giving the solution to that:

```
Caused by:
  feature `sentinel` includes `dep:tokio/sync` with both `dep:` and `/`
  To fix this, remove the `dep:` prefix.
  If the intent is to avoid creating an implicit feature `tokio` for an optional dependency, then consider replacing this with two values:
      "dep:tokio", "tokio/sync"
```

This works and no longer exposes `tokio` as a feature:

```diff
-sentinel = ["redis/sentinel", "tokio/sync"]
+sentinel = ["redis/sentinel", "dep:tokio", "tokio/sync"]
```

Alternatively we could change it like that which does the exact same thing:

```diff
-sentinel = ["redis/sentinel", "tokio/sync"]
+sentinel = ["redis/sentinel", "dep:tokio"]
...
-tokio = { version = "1.0", default-features = false, optional = true }
+tokio = { version = "1.0", default-features = false, features = ["sync"], optional = true }
```

:partying_face:

bikeshedder:
If you do make `cluster` an alias to `cluster-async` you also need to make sure that the `#[cfg(feature = "cluster")]` also is changed.

Personally I'd make `cluster-async` an alias to `cluster` and leave the code as is. This would make it identical to the `sentinel` feature which doesn't have an `-async` variant to begin with. As long as one is an alias of the other and both behave identical it doesn't really matter. I leave that up to you. :sweat_smile:

bikeshedder:
Is this PR ready to be merged or are you still working on it?

woshilapin:
OK, I've put back `cluster` feature as it was, and made `cluster-async` an alias to `cluster`.
I've also added the `dep:tokio` to `sentinel` feature.
See the changes [here](https://github.com/deadpool-rs/deadpool/compare/c20e66bcf5bf9586f31d4ce24733f56e612ac47e..57f600bf1283e639b2e208520d668d8892b3f119).

I believe the PR is ready to be merged. It would probably take another round later to extend that to all other crates of `deadpool` in the same way.

redis/README.md
```diff
@@ -14,6 +14,10 @@ manager for [`redis`](https://crates.io/crates/redis).
 | `rt_async-std_1` | Enable support for [async-std](https://crates.io/crates/config) crate | `deadpool/rt_async-std_1`, `redis/async-std-comp` | no      |
 | `serde`          | Enable support for [serde](https://crates.io/crates/serde) crate      | `deadpool/serde`, `serde/derive`                  | no      |
 | `cluster`        | Enable support for Redis Cluster                                      | `redis/cluster-async`                             | no      |
+| `sentinel`       | Enable high-level interfaces for communication with Redis sentinels   | `redis/sentinel`, `tokio/sync`                    | no      |
+
+All of the features of [redis](https://crates.io/crates/redis) are also reexposed with a `redis-` prefix.
```
bikeshedder, line 19: Since we dropped the prefix this isn't true anymore.

bikeshedder:
Great! Just two nit picks and I'll merge it asap. :grinning: :+1: 

Thanks a lot for the work on this!

bikeshedder:
I just released `deadpool-redis` version `0.20.0` which includes this change:

:package: https://crates.io/crates/deadpool-redis/0.20.0

PR #391, opened 2025-02-19
Expose features

### LINKED ISSUE #169: Re-export features of dependencies

woshilapin:
Follow-up of #389 and should fix #169.

🔍 This PR can be reviewed commit by commit.

The first commit make a more generic version of all the scripts that will check if features are re-exposed.

> [!IMPORTANT]
> I did reexposed only the crates that re-export publicly crates as subcrates (e.g. `pub use redis`). Hence I did not do the work for `memcached`.


bikeshedder:
`deadpool-memcached` is barely ever used and it's in a broken state right now. I looked into that a few days ago but still have to finish that PR:

- #368

bikeshedder:
This is awesome! :partying_face:

I really have to learn `jqlang`. I probably would've written a bunch of python script or even some rust code instead. I remember reading about `jq` some time ago but didn't really see any use case for it until now. :+1:

bikeshedder:
Fyi. The `matrix` must be an object rather than a sequence. I fixed it by adding `include` as key and putting the sequence there: https://github.com/deadpool-rs/deadpool/commit/e384850eb02c3d9ad4b5912df6a36b64d249deeb

PR #398, opened 2025-03-19
Iam

emcee21:


bikeshedder:
I superficially reviewed it and there is one thing that I really don't like about it:

Rather than adding new methods to the `ManagerConfig` and adding new fields to the `ManagerConfig` I would prefer if this was designed more like a plug-in. I'd like to see a `ConfigProvider` trait that can then be made as AWS specific as necessary:

e.g.
```rust
pub trait ConfigProvider {
    async fn get_config(&self) -> Config;
}
```

I'm really just sketching this in my head right now.

Maybe `Config` should be made a trait and `BasicConfig` and `AwsConfig` could then implement this new trait?

If connections need to be discarded after a given time that could be done via a `pre_recycle` hook.

emcee21:
Makes sense. Fwiw I thought the “max_age” feature might have applicability beyond AWS.

bikeshedder:
The `max_age` feature is something that can be of interest for every `deadpool-*` crate. There are some examples how to write that max age hook in the repo. It really needs some better documentation though.

bikeshedder:
Found the example:
- https://github.com/deadpool-rs/deadpool/issues/359#issuecomment-2402643536

```rust
Pool::builder(...)
    // ...
    .pre_recycle(Hook::sync_fn(|_, metrics| {
        if metrics.age() > max_age {
            Err(HookError::message("Max age reached"))
        } else {
            Ok(())
        }
    }))
    // ...
```

bikeshedder:
It'd be nice if this felt more like a plugin. I've been toying with the idea of plugins some time ago. The API would then look like that:

```rust
Pool::builder(...)
    .add_plugin(MaxAge(Duration::from_seconds(300))
```

There are some unanswered questions though:
- How does one change the configuration of a plugin?
- How to add and remove plugins while the pool is running?
- The order of plugins does matter! (Not a question but a finding while working on that idea)

Those were the reasons why I think making the entire pool hot swappable is the way to go in the future. This would allow maximum flexibility and enable basically every nice use case that I can think of.

---

The auth credentials are a bit different though. They don't require a pool reconfiguration and swapping of the manager. I can imagine some provider using OTPs that need to change for every single connection. So adding a way to dynamically create the configuration as needed makes a lot of sense.

emcee21:
No worries at all! My plan is to keep debugging the current implementation. As you gather your thoughts I can come back to implement this PR as a proper candidate.

bikeshedder:
- #219

PR #401, opened 2025-04-09
Fix Wikipedia link

jasonaowen:
Reference-style Markdown links can use angle brackets or nothing, but not parentheses.

https://www.markdownguide.org/basic-syntax/#formatting-the-second-part-of-the-link


bikeshedder:
Thank you! :heart:

PR #417, opened 2025-05-20
Build(Redis): Add serde dependency to sentinel feature

MuhtasimTanmoy:
Fix https://github.com/deadpool-rs/deadpool/issues/416


bikeshedder:
`serde` should be optional and not automatically enabled with the `sentinel` feature. I'll push a fix for that in a sec.

MuhtasimTanmoy:
Hi, as `sentinel` feature has explicit imports and usage of `serde`, is it possible for `serde` to be optional?

https://github.com/deadpool-rs/deadpool/blob/35d158c43669db8e16fce4bb92f2b3329f9104c2/redis/src/sentinel/config.rs#L3


https://github.com/deadpool-rs/deadpool/blob/35d158c43669db8e16fce4bb92f2b3329f9104c2/redis/src/sentinel/config.rs#L51

MuhtasimTanmoy:
In comparison to `cluster` feature, there `serde` is fully optional. But not in `sentinel`.

https://github.com/deadpool-rs/deadpool/blob/35d158c43669db8e16fce4bb92f2b3329f9104c2/redis/src/cluster/config.rs#L40


Causing https://github.com/deadpool-rs/deadpool/issues/416 and https://github.com/deadpool-rs/deadpool/issues/414

MuhtasimTanmoy:
Ok.

PR #419, opened 2025-06-08
Fix comment for PoolInner::users

rjuju:
Oversight in 7ffa964 which renamed `PoolInner::available` to `PoolInner::users`.

### DIFF
--- a/src/managed/mod.rs
+++ b/src/managed/mod.rs
@@ -594,9 +594,8 @@ impl<M: Manager, W: From<Object<M>>> Pool<M, W> {
 struct PoolInner<M: Manager> {
     manager: M,
     slots: Mutex<Slots<ObjectInner<M>>>,
-    /// Number of available [`Object`]s in the [`Pool`]. If there are no
-    /// [`Object`]s in the [`Pool`] this number can become negative and store
-    /// the number of [`Future`]s waiting for an [`Object`].
+    /// Number of ['Pool'] users. A user is both a future which is waiting for an ['Object'] or one
+    /// with an ['Object'] which hasn't been returned, yet.
     users: AtomicUsize,
     semaphore: Semaphore,
     config: PoolConfig,


src/managed/mod.rs
```diff
@@ -594,9 +594,7 @@ impl<M: Manager, W: From<Object<M>>> Pool<M, W> {
 struct PoolInner<M: Manager> {
     manager: M,
     slots: Mutex<Slots<ObjectInner<M>>>,
-    /// Number of available [`Object`]s in the [`Pool`]. If there are no
-    /// [`Object`]s in the [`Pool`] this number can become negative and store
-    /// the number of [`Future`]s waiting for an [`Object`].
+    /// Number of [`Object`] currently used in the [`Pool`].
```
bikeshedder, line 597: That's not the correct description of the `users` field. It is the number of objects currently requested from the pool including futures who got their objects from the pool and futures waiting for their turn.

You can also see that in the implementation of `status()`:
```rust
let (available, waiting) = if users < slots.size {
    (slots.size - users, 0)
} else {
    (0, users - slots.size)
};
```

I was told that the `users` metric is not very useful so it became two values `available` and `waiting` in the public API which is much easier to reason about.

bikeshedder: The current comment is wrong though. It can never become negative. It used to (when it was still called `available`) but that caused a lot of confusion and the `users` count is much more intuitive. :sweat_smile:

If you got a good idea how to describe this field go ahead. Otherwise I'll just change the comment to:
```rs
/// Number of pool users. A user is both a future which is waiting for an
/// object or one with an object which hasn't been returned, yet.
```

Does that sound right to you?

rjuju: And agreed, the users count is more intuitive :)

bikeshedder:
Thanks a lot for bringing this to my attention and creating a PR for it!. :heart:

PR #424, opened 2025-06-22
Libsql integration

Nerglej:
Hi!

I've made a simple libsql integration, that I already use in a project. Since libsql, as default, is async, I've only implemented the async traits. And because of how the connections are made in libsql, the integrations doesn't even need a config (inspired by the memchached integration).

I would also like to note that I've never made a changelog file in this format before, so if someone could check that and correct it, that would be amazing.

Any other notes or changes you see fit?

Best regards,
William


bikeshedder:
> Because of how connections are made in libsql, the integrations doesn't really need a config (inspired by the memchached integration).

While `libsql` doesn't have a classic connection structure they use the `Builder` pattern. e.g.
`Builder::new_local` is equal to a local conneaction configuration. I would expect a config such as:

```rs
struct Config {
    database: DatabaseVariant,
}

enum DatabaseVariant {
    Local { path: Path },
    RemoteReplica { path: Path, url: String, auth_token: String },
    LocalReplica { path: Path },
    SyncedDatabase { path: Path, url: String, auth_token: String },
    Remote { url: String, auth_token: String }
}
```

It's a bit more involved than that as there are plenty of other options such as `flags`, `encryption_config`, `skip_safety_assert`, etc. which depend on the database variant.

This is a kind of configuration. They just used the builder pattern for it and called the struct `Builder`.

:+1:  Not having a configuration structure isn't a blocker. It's just a nice to have.

Nerglej:
I see that a config would be nice.

I would argue, though, that using as much from the integrated library as possible would be
better, for future maintanance. Especially in the case of a database as "feature-rich"
as sqlite and libsql.

I see that libsql use a `Builder` pattern, but even after you've called the build-function
you still need to create a connection by calling `.connect()`. I saw the libsql
database object as the config itself, which would lessen the burden on this library,
and at the same time gives users the option to do whatever libsql allows with
their configurations.

I see that both a Config structure and the currently implemented way is possible
to have at the same time, but again, I would argue that a config would just be
a wrapper around another thing that does mostly the same thing, but with limits.

However, I see while writing this, that I could implement it kind of like in the
Sqlite implementation. I'll try to implement a way of mixing both to make the DX
the best as to my ability, with as little limitations as possible.

Best regards,
William

Nerglej:
ae20906 is an attempt to do both. The downside currently is that we can't serialize and deserialize the database provided from libsql.

It simplified the DX a bit, since there's just "two" entrypoints, the libsql database and the config, where we before had both the libsql database, the deadpool manager and the deadpool pool.

Again, to argue against implementing the whole thing, is that people can look to libsql for documentation, instead of here.

Any comments or tips?

bikeshedder:
I just merged #433 which renders this PR redundant.

The code that got merged has an almost complete config structure except for the `connector` and `http_request_callback` parameters which I don't consider runtime configuration anyways.

The crate also got released on crates.io:

:package: https://crates.io/crates/deadpool-libsql

PR #451, opened 2025-10-29
Debug and doc fixes

werekraken:


crates/deadpool/src/managed/pool.rs
```diff
@@ -416,8 +416,8 @@ pub(crate) struct PoolInner<M: Manager> {
     manager: M,
     next_id: AtomicUsize,
     slots: Mutex<Slots<ObjectInner<M>>>,
-    /// Number of ['Pool'] users. A user is both a future which is waiting for an ['Object'] or one
-    /// with an ['Object'] which hasn't been returned, yet.
+    /// Number of [`Pool`] users. A user is any future that has acquired a semaphore permit,
+    /// including both those waiting for an [`Object`] and those currently holding an [`Object`].
```
bikeshedder: I remember rewriting this sentence over and over again. I tried incorporating the idea of "borrowing objects from the pool" but couldn't come up with a great phrasing.

crates/deadpool/src/managed/pool.rs
```diff
@@ -416,8 +416,8 @@ pub(crate) struct PoolInner<M: Manager> {
     manager: M,
     next_id: AtomicUsize,
     slots: Mutex<Slots<ObjectInner<M>>>,
-    /// Number of ['Pool'] users. A user is both a future which is waiting for an ['Object'] or one
-    /// with an ['Object'] which hasn't been returned, yet.
+    /// Number of [`Pool`] users. A user is any future that has acquired a semaphore permit,
```
bikeshedder: A user is both a future that is waiting for a semaphore permit and a future that has acquired a permit. Essentially the semaphore is the first line of defense to borrow an object from the pool. If a future has acquired a permit it is allowed to either reuse an object stored in the `slots` vector or create a new one. A future that hasn't acquired a future is considered "waiting".

bikeshedder (changes requested):
Thanks a lot.

crates/deadpool/src/managed/pool.rs
werekraken: Dropping for now--if I come up with anything better I'll submit it as a separate PR.

werekraken: What do you think of this instead?
```
-    /// Number of [`Pool`] users. A user is both a future which is waiting for an [`Object`] or one
-    /// with an [`Object`] which hasn't been returned, yet.
+    /// Number of [`Pool`] users. A user is a future, including both those waiting for an [`Object`]
+    /// and those currently holding an [`Object`].
```
If you like it, I'm happy to add it to this PR or submit it as a followup PR.

bikeshedder:
Thanks a lot! :partying_face:

PR #457, opened 2025-12-17
deadpool-redis: Re-export the `uuid` feature of the `redis` crate.

cdbfoster:
The [docs](https://docs.rs/deadpool-redis/latest/deadpool_redis/) state that:

    All of the features of [redis](https://crates.io/crates/redis) are also re-exported.

Seem to have missed this one, or the docs are inaccurate.


bikeshedder:
This happens when a minor version update adds a feature. The `check-reexports` check of the CI catches that but that doesn't happen automatically when a new `redis` release is made. A cron of some kind would be nice to run those checks when a new version of the dependencies is published.

PR #461, opened 2026-01-30
fix: Fix a panic when calling `JoinError::into_panic`

Hywan:
That's a lot of panics. This patch fixes a panic raised by `JoinError:into_panic` when the `JoinError` represents a cancellation instead of a panic.

---

* Fix https://github.com/deadpool-rs/deadpool/issues/460


bikeshedder:
> https://github.com/deadpool-rs/deadpool/releases/tag/deadpool-runtime-v0.3.1 apparently it's here

Almost. The runtime crate got updated but as it is a breaking change a release of the other `deadpool-*` crates is required, too. I'm preparing a release of the entire `deadpool-*` suite. I'll keep you updated when that's been done.

PR #466, opened 2026-02-18
Fix flaky cluster `test_recycled` by using pool object ID

### LINKED ISSUE #465: redis cluster test `test_recycled` fails sometimes
The `test_recycled` of the redis cluster code fails sometimes inside the CI:

See: https://github.com/deadpool-rs/deadpool/actions/runs/22133300586/job/63978921634

bombsimon:
`CLIENT SETNAME`/`GETNAME` are per-TCP-connection commands. In a cluster, `ClusterConnection` manages multiple internal TCP connections (one per node). redis-rs routes `CLIENT SETNAME` to all nodes but `CLIENT GETNAME` to a random single node, so `GETNAME` can return `None` if it hits a different node than expected.

Instead, use deadpool's `Object::id()` to verify recycling at the pool level. This is deterministic and doesn't depend on Redis cluster command routing.

Also expose `id()` and `metrics()` on the cluster `Connection` wrapper.

Fixes #465

### DIFF
--- a/crates/deadpool-redis/src/cluster/mod.rs
+++ b/crates/deadpool-redis/src/cluster/mod.rs
@@ -45,12 +45,12 @@ impl Connection {
     }
 
     /// Returns the unique [`ObjectId`] of this [`Connection`].
-    pub fn id(this: &Self) -> deadpool::managed::ObjectId {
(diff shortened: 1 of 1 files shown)


bikeshedder:
Awesome. Thanks a lot! :heart:

PR #474, opened 2026-03-10
Add AsyncConnectionConfig support to Manager

bombsimon:
Add `Manager::new_with_config` which accepts an `AsyncConnectionConfig`, allowing connection-level settings (e.g. response timeout, connection timeout) to be configured when creating pooled connections.

The existing `Manager::new` behavior is unchanged. When a config is provided, `get_multiplexed_async_connection_with_config` is used instead of `get_multiplexed_async_connection`.

Relates to #452 (you can set timeout on `AsyncConnectionConfig`)

---

Before Redis 1 we only get a response timeout [when it was explicitly set](https://github.com/redis-rs/redis-rs/blob/193f51771e545e9ade61cf756b269511af5de046/redis/src/client.rs#L205-L209), probably via any of the constructors such as [`get_multiplexed_async_connection_with_timeouts`](https://github.com/redis-rs/redis-rs/blob/193f51771e545e9ade61cf756b269511af5de046/redis/src/client.rs#L307)

With version 1, we now get a [default of 500ms](https://github.com/redis-rs/redis-rs/blob/3ebba95c86417a1c36c7881866ed427e74feda76/redis/src/client.rs#L178) if no `AsyncConnectionConfig` is passed.

Since we can't pass this through `deadpool-redis`, we'll always end up with an initial 500ms and to change that we'll do something like

```rust
let mut conn = pool.get().await?;
conn.set_response_timeout(Duration::from_secs(300));
```

That's also not fully compatible because `set_response_timeout` still doesn't accept `None` which is a valid value.

### DIFF
--- a/crates/deadpool-redis/src/lib.rs
+++ b/crates/deadpool-redis/src/lib.rs
@@ -35,7 +35,7 @@ use std::{
 
 use deadpool::managed;
 use redis::{
-    Client, IntoConnectionInfo, RedisError, RedisResult,
+    AsyncConnectionConfig, Client, IntoConnectionInfo, RedisError, RedisResult,
     aio::{ConnectionLike, MultiplexedConnection},
 };
 
@@ -127,10 +127,19 @@ impl ConnectionLike for Connection {
 /// [`Manager`] for creating and recycling [`redis`] connections.
 ///
 /// [`Manager`]: managed::Manager
-#[derive(Debug)]
 pub struct Manager {
     client: Client,
+    connection_config: Option<AsyncConnectionConfig>,
     ping_number: AtomicUsize,
+}
+
+impl std::fmt::Debug for Manager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Manager")
+            .field("client", &self.client)
+            .field("ping_number", &self.ping_number)
+            .finish()
+    }
 }
 
 impl Manager {
@@ -142,6 +151,27 @@ impl Manager {
     pub fn new<T: IntoConnectionInfo>(params: T) -> RedisResult<Self> {
         Ok(Self {
             client: Client::open(params)?,
+            connection_config: None,
+            ping_number: AtomicUsize::new(0),
+        })
+    }
+
+    /// Creates a new [`Manager`] from the given `params` and
+    /// [`AsyncConnectionConfig`].
+    ///
+    /// This allows configuring connection-level settings such as response timeouts and connection
+    /// timeouts.
+    ///
+    /// # Errors
+    ///
+    /// If establishing a new [`Client`] fails.
+    pub fn new_with_config<T: IntoConnectionInfo>(
+        params: T,
+        connection_config: AsyncConnectionConfig,
+    ) -> RedisResult<Self> {
+        Ok(Self {
+            client: Client::open(params)?,
+            connection_config: Some(connection_config),
             ping_number: AtomicUsize::new(0),
         })
     }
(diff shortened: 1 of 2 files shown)


crates/deadpool-redis/src/lib.rs
```diff
@@ -127,12 +127,21 @@ impl ConnectionLike for Connection {
...
 }
 
+impl std::fmt::Debug for Manager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Manager")
+            .field("client", &self.client)
+            .field("ping_number", &self.ping_number)
+            .finish()
+    }
+}
```
bombsimon, line 143: Since `Manager` used to derive `Debug`, dropping that would be breaking and since `AsyncConnectionConfig` doesn't impl `Debug` I added this which will result in the same as today.

crates/deadpool-redis/tests/redis_connection_config.rs
```diff
@@ -0,0 +1,89 @@
...
+#[tokio::test]
+async fn test_response_timeout_can_be_disabled() {
+    let config = AsyncConnectionConfig::new().set_response_timeout(None);
+    let pool = create_pool_with_config(config);
+    let mut conn = pool.get().await.unwrap();
+
+    let result: Option<(String, String)> = conn
+        .blpop("deadpool/nonexistent_timeout_test_key", 1.0)
+        .await
+        .unwrap();
```
bombsimon, line 63: Maybe not ideal to slow down CI with 1s but couldn't come up with a better way to assert this. Maybe there are other things we can test from `AsyncConnectionConfig` but used this since the issue I had was from timeouts.

bombsimon:
Re-exports not really related to this, should I add them in this PR or a separate one and rebase after merge? Or something else?

<details>
<summary>Diff</summary>

```diff
diff --git a/crates/deadpool-redis/Cargo.toml b/crates/deadpool-redis/Cargo.toml
index 98ac2db..aced82c 100644
--- a/crates/deadpool-redis/Cargo.toml
+++ b/crates/deadpool-redis/Cargo.toml
@@ -31,6 +31,7 @@ cache-aio = ["redis/cache-aio"]
 cluster = ["redis/cluster-async"]
 cluster-async = ["cluster"]
 connection-manager = ["redis/connection-manager"]
+entra-id = ["redis/entra-id"]
 geospatial = ["redis/geospatial"]
 json = ["redis/json"]
 num-bigint = ["redis/num-bigint"]
@@ -41,6 +42,7 @@ smol-comp = ["redis/smol-comp"]
 smol-native-tls-comp = ["redis/smol-native-tls-comp"]
 smol-rustls-comp = ["redis/smol-rustls-comp"]
 streams = ["redis/streams"]
+token-based-authentication = ["redis/token-based-authentication"]
 tls-native-tls = ["redis/tls-native-tls"]
 tls-rustls = ["redis/tls-rustls"]
 tls-rustls-insecure = ["redis/tls-rustls-insecure"]
```
</details>

bikeshedder:
Ah, I merged that a bit too soon without thinking about that a lot.

While it makes sense to configure the timeouts with `deadpool-redis` things like `push_sender` are rather useless as a configuration option and should be managed by deadpool itself.

It would actually be nice if the timeouts were part of the `config` structures rather than exposing the `AsyncConnectionInfo` directly.

Sadly the `redis` crate has no proper separation between those two types of configuration:
- config a user of an application should be able to provide
- config parameters required by the application to run

Is there anything except for the timeouts you're currently interested in?

bombsimon:
Yeah that makes sense, sorry for opening a PR without issue first to discuss. 

For now, all I need are the timeouts. I can do a followup PR to just expose those instead tomorrow and not the whole type if you want to?

bikeshedder:
> For now, all I need are the timeouts. I can do a followup PR to just expose those instead tomorrow and not the whole type if you want to?

That'd be awesome! I'm currently quite limited with my time and just thought that I quickly merge this good looking PR which even included tests (:partying_face:). Just to realize a few moments later that I didn't check what else the `AsyncConnectionConfig` includes. And then I remembered looking at that a few weeks ago. :facepalm:

PR #475, opened 2026-03-12
Replace AsyncConnectionConfig with Manager builder pattern

bombsimon:
Removes `Manager::new_with_config` which exposed internal redis settings. Adds `Manager::builder()` with `connection_timeout()` and `response_timeout()` setters that only surface user-facing options.

---

Followup from #474. Thanks for the quick turnaround and sorry for confusing you and just exposing the whole config.

Since this now adds two new settings to the `Manager` and there's already a builder pattern for the `Config` I opted to use the builder pattern here as well. LMK if you rather have a separate `new_with_timeouts` constructor or something instead.

### DIFF
--- a/crates/deadpool-redis/src/config.rs
+++ b/crates/deadpool-redis/src/config.rs
@@ -1,4 +1,4 @@
-use std::{fmt, path::PathBuf};
+use std::{fmt, path::PathBuf, time::Duration};
 
 use redis::RedisError;
 #[cfg(feature = "serde")]
@@ -17,6 +17,10 @@ use crate::{CreatePoolError, Pool, PoolBuilder, PoolConfig, RedisResult, Runtime
 /// REDIS__POOL__MAX_SIZE=16
 /// REDIS__POOL__TIMEOUTS__WAIT__SECS=2
 /// REDIS__POOL__TIMEOUTS__WAIT__NANOS=0
+/// REDIS__CONNECTION_TIMEOUT__SECS=5
+/// REDIS__CONNECTION_TIMEOUT__NANOS=0
+/// REDIS__RESPONSE_TIMEOUT__SECS=2
+/// REDIS__RESPONSE_TIMEOUT__NANOS=0
 /// ```
 /// ```rust
 /// #[derive(serde::Deserialize)]
@@ -47,6 +51,22 @@ pub struct Config {
 
     /// Pool configuration.
     pub pool: Option<PoolConfig>,
+
+    /// Connection timeout applied when creating new Redis connections.
+    ///
+    /// - `None`: use the redis crate's default.
+    /// - `Some(None)`: disable the timeout.
+    /// - `Some(Some(duration))`: use this specific timeout.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub connection_timeout: Option<Option<Duration>>,
+
+    /// Response timeout applied when waiting for Redis responses.
+    ///
+    /// - `None`: use the redis crate's default.
+    /// - `Some(None)`: disable the timeout.
+    /// - `Some(Some(duration))`: use this specific timeout.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub response_timeout: Option<Option<Duration>>,
 }
 
 impl Config {
@@ -70,13 +90,36 @@ impl Config {
     /// See [`ConfigError`] for details.
     pub fn builder(&self) -> Result<PoolBuilder, ConfigError> {
         let manager = match (&self.url, &self.connection) {
-            (Some(url), None) => crate::Manager::new(url.as_str())?,
-            (None, Some(connection)) => crate::Manager::new(connection.clone())?,
-            (None, None) => crate::Manager::new(ConnectionInfo::default())?,
+            (Some(url), None) => self.build_manager(url.as_str())?,
+            (None, Some(connection)) => self.build_manager(connection.clone())?,
+            (None, None) => self.build_manager(ConnectionInfo::default())?,
             (Some(_), Some(_)) => return Err(ConfigError::UrlAndConnectionSpecified),
         };
+
         let pool_config = self.get_pool_config();
+
         Ok(Pool::builder(manager).config(pool_config))
+    }
+
+    fn build_manager<T: redis::IntoConnectionInfo>(
+        &self,
+        params: T,
+    ) -> Result<crate::Manager, ConfigError> {
+        if self.connection_timeout.is_some() || self.response_timeout.is_some() {
+            let mut config = redis::AsyncConnectionConfig::new();
+
+            if let Some(timeout) = self.connection_timeout {
+                config = config.set_connection_timeout(timeout);
+            }
+
+            if let Some(timeout) = self.response_timeout {
+                config = config.set_response_timeout(timeout);
+            }
+
+            Ok(crate::Manager::new_with_config(params, config)?)
+        } else {
+            Ok(crate::Manager::new(params)?)
+        }
     }
 
     /// Returns [`deadpool::managed::PoolConfig`] which can be used to construct
@@ -94,6 +137,8 @@ impl Config {
             url: Some(url.into()),
             connection: None,
             pool: None,
+            connection_timeout: None,
+            response_timeout: None,
         }
     }
 
@@ -105,7 +150,29 @@ impl Config {
             url: None,
             connection: Some(connection_info.into()),
             pool: None,
+            connection_timeout: None,
+            response_timeout: None,
         }
+    }
+
+    /// Sets the connection timeout.
+    ///
+    /// Pass `Some(duration)` to set a specific timeout, or `None` to
+    /// explicitly disable it.
+    #[must_use]
+    pub fn with_connection_timeout(mut self, timeout: Option<Duration>) -> Self {
+        self.connection_timeout = Some(timeout);
+        self
+    }
+
+    /// Sets the response timeout.
+    ///
+    /// Pass `Some(duration)` to set a specific timeout, or `None` to
+    /// explicitly disable it.
+    #[must_use]
+    pub fn with_response_timeout(mut self, timeout: Option<Duration>) -> Self {
+        self.response_timeout = Some(timeout);
+        self
     }
 }
 
@@ -115,6 +182,8 @@ impl Default for Config {
             url: None,
             connection: Some(ConnectionInfo::default()),
             pool: None,
+            connection_timeout: None,
+            response_timeout: None,
         }
     }
 }
--- a/crates/deadpool-redis/src/config.rs
+++ b/crates/deadpool-redis/src/config.rs
@@ -6,6 +6,9 @@ use serde::{Deserialize, Serialize};
 
 use crate::{CreatePoolError, Pool, PoolBuilder, PoolConfig, RedisResult, Runtime};
 
+const DEFAULT_CONNECTION_TIMEOUT: Option<Duration> = Some(Duration::from_secs(1));
+const DEFAULT_RESPONSE_TIMEOUT: Option<Duration> = Some(Duration::from_millis(500));
+
 /// Configuration object.
 ///
 /// # Example (from environment)
@@ -17,10 +20,10 @@ use crate::{CreatePoolError, Pool, PoolBuilder, PoolConfig, RedisResult, Runtime
 /// REDIS__POOL__MAX_SIZE=16
 /// REDIS__POOL__TIMEOUTS__WAIT__SECS=2
 /// REDIS__POOL__TIMEOUTS__WAIT__NANOS=0
-/// REDIS__CONNECTION_TIMEOUT__SECS=5
+/// REDIS__CONNECTION_TIMEOUT__SECS=1
 /// REDIS__CONNECTION_TIMEOUT__NANOS=0
-/// REDIS__RESPONSE_TIMEOUT__SECS=2
-/// REDIS__RESPONSE_TIMEOUT__NANOS=0
+/// REDIS__RESPONSE_TIMEOUT__SECS=0
+/// REDIS__RESPONSE_TIMEOUT__NANOS=500000000
 /// ```
 /// ```rust
 /// #[derive(serde::Deserialize)]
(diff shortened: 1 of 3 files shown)


bikeshedder:
That's much better. Though I'm seriously wondering if we could add a proper configuration object instead. What I mean by that is the following: Almost all `deadpool-*` crates maintained by me try to offer a ready-to-use configuration object that can be dropped into your own configuration. That makes the usage in actual applications much simpler as you don't need to worry about defining your own configuration structure for database connectivity.

https://github.com/deadpool-rs/deadpool/blob/main/crates/deadpool-redis/src/config.rs

For that reason all config structures in deadpool are serde (de)serializable.

bombsimon:
> Though I'm seriously wondering if we could add a proper configuration object instead. 

SGTM! I updated the code to now put the timeouts on the existing `Config` struct. I added to `with_` setters similar to `sentinel::Config`.

With this I now dropped the builder for `Manager` and our config will build it instead. A result is that you can't build a `Manager` directly with timeouts without going via the `Config`. I'm not sure this is an issue? If so, should I create a new path where we can build a manager by passing timeouts and/or just use that one instead of `build_manager` on `Config`? I avoided it because it doesn't scale that well adding new constructors for every new config permutation.

crates/deadpool-redis/src/lib.rs
```diff
@@ -145,81 +151,18 @@ impl Manager {
...
-    ///
-    /// # Errors
-    ///
-    /// If establishing a new [`Client`] fails.
-    pub fn build(self) -> RedisResult<Manager> {
-        Ok(Manager {
-            client: Client::open(self.params)?,
-            connection_timeout: self.connection_timeout,
-            response_timeout: self.response_timeout,
+    fn new_with_config<T: IntoConnectionInfo>(
```
bikeshedder: Now that I think about it I think we should keep that method public. Otherwise users can only create a manager with timeouts using the config way.

I'm unsure if we should force users to always use the config code path. Even though I find that much more ergonomic than fiddling with the `redis` structures directly.

crates/deadpool-redis/src/config.rs
```diff
@@ -47,6 +51,22 @@ pub struct Config {
...
     /// Pool configuration.
     pub pool: Option<PoolConfig>,
+
+    /// Connection timeout applied when creating new Redis connections.
+    ///
+    /// - `None`: use the redis crate's default.
+    /// - `Some(None)`: disable the timeout.
+    /// - `Some(Some(duration))`: use this specific timeout.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub connection_timeout: Option<Option<Duration>>,
```
bikeshedder: That `Option<Option<Duration>>` caught me off guard. I doubt that `serde` understands how to deserialize that. It's very confusing and the sheer necessity of the comment explaining the behavior is a red flag to me.

I'd prefer if we changed this into an `Option<Duration>` which uses a custom default function that returns `Some(DEFAULT_CONNECTION_TIMEOUT)` instead.

The `redis` crate defines them as following:
```rs
#[cfg(feature = "aio")]
pub(crate) const DEFAULT_RESPONSE_TIMEOUT: Option<Duration> = Some(Duration::from_millis(500));
#[cfg(any(feature = "aio", feature = "cluster"))]
pub(crate) const DEFAULT_CONNECTION_TIMEOUT: Option<Duration> = Some(Duration::from_secs(1));
```

Sadly it's not exported as `pub` so we can't reuse that definition in `deadpool-redis`. We should be safe to copy those values however. I hope that they'll never change.

bikeshedder (changes requested):
Just a few nitpicks. `Option<Option<Duration>>` really isn't great DX and I think we should expose `new_with_config` as public method.

crates/deadpool-redis/src/lib.rs
bombsimon: Yeah that was my concern as well above and I agree it's not ideal. What do you want this method to accept? Two durations? A struct containing them? A builder pattern?

bikeshedder: I'd just go with `Option<Duration>` and implement `Default` with the timeouts set to `Some(DEFAULT_CONNECTION_TIMEOUT)` and `Some(DEFAULT_RESPONSE_TIMEOUT)`. For the serde part add `default_connection_timeout` and `default_response_timeout` functions that returns those values and put them in the `#[serde(default=...)]`.

crates/deadpool-redis/src/config.rs
bombsimon: Sorry, my bad. I was under the impression that `#[serde(default)]` would differentiate between `{"key":null}` and just omitting the key but I was forgetting `skip_serializing_if = "Option::is_none")`. However seems that still didn't help 🤔 

The reason I used `Option<Option<T>>` is because not passing a timeout will use the default timeout whereas passing `None` will use _no_ timeout.

- `None` -> Use upstream default (currently 500ms)
- `Some(None)` -> Pass `None` to the config, disabling timeout
- `Some(Some(Duration))` -> Pass `Duration` as timeout

If we use an `Option<T>` we won't be able to ever keep the `redis` upstream default and always use our own. Is that ok?

bikeshedder: Yes. That's fine for now. Ideally we could get the `redis` crate to expose those defaults public constants but for the time being it's better to define our own. It is a much cleaner API.

bombsimon: So basically this was what I was aiming for which can differentiate between them: [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=64d0422f9d1ce8fed44ad07b3339fdb8)

EDIT: Oops, dropped connection between your response. Will skip custom deserialization and use `Option<T>` for now. Sorry for the noise!

crates/deadpool-redis/src/lib.rs
bikeshedder: I see your point. The `AsyncConnectionConfig` is essentially a "write-only" structure. I hate this pattern and I don't get why so many libraries do that. 😫 

I think there are basically two options. Roll our own structure - which is basically what you did at first - or accept that this is essentially something opaque passed by the user and we have no control over it. We can't even validate it.

The only real trouble maker I see in this struct is `set_push_sender` which deadpool will have to override in the future in order to support RESP3.

crates/deadpool-redis/src/config.rs
bombsimon: And yeah looks safe to just copy over the constants. According to https://github.com/redis-rs/redis-rs/issues/1932#issuecomment-3801568742 changing them is a breaking change they don't want to do and if they plan to follow proper semver this will require a v2 to change.

crates/deadpool-redis/src/lib.rs
bombsimon: I think a builder pattern (the `ManagerBuilder`) makes most sense because it's forward compatible and will allow you to add new config options without breaking backwards compatibility.

Then also having the values on the `Config`, defaulting to the same as `redis`, and pass them to the manager builder pattern.

bikeshedder: A `struct` with `#[non_exhaustive]` works just as well for forward compatibility and supports more use cases. In my opinion the builder pattern is highly overrated and used way too much in APIs. A plain struct with public fields often does the job just as well, requires less code and reads better 90% of the time.

The main exception is when there is an invariant the type must enforce, for example when two fields are mutually exclusive. But even then, that is often better modeled as an enum than as a builder.

The builder pattern has its place, but for configuration options I personally dislike it, especially when it is used in a write-only style for no good reason.

bikeshedder: Sorry for getting philosophical - I've often heard the argument that those setter and getter interfaces are for decoupling the implementation from the interface. I've yet to come across an actual situation where an internal refactor is possible without affecting the public interface. It sounds great in theory but typically such things result in a breaking change anyways. I rather make a breaking release than offering an outdated and possibly misleading interface. :see_no_evil:

bombsimon: Don't be sorry, appreciate the discussions and that you take the time to get back to me! I just hope you don't see this as more work than fixing it yourself. I have no strong argument for using the builder pattern!

I generally don't mind breaking changes but given so many Rust crates never reach 1.0 and constantly break code in minor releases that can be a bit cumbersome to work with.

To summarize, creating a new struct with config for the manager, adding `#[non_exhaustive]` and dropping the builder is the desired way then? Do you want me to drop `with_connection_timeout` and `with_response_timeout` from the `Config` as well?

**EIDT**: I'm pushing some changes between replies mostly to speed up turnaround and confirm changes are on track, LMK if that not desired. I can squash later or you when merging to reduce commits.

bikeshedder: tbh. I was about to make the point that deadpool uses plain structs everywhere just to find that this is not necessarily true...
- https://docs.rs/deadpool/latest/deadpool/managed/struct.PoolConfig.html
- https://docs.rs/deadpool/latest/deadpool/managed/struct.PoolBuilder.html
- https://docs.rs/deadpool-postgres/latest/deadpool_postgres/struct.Config.html

I do feel like the `PoolBuilder` is the actual outlier here and only exists because of an improved DX for the hooks. I do wonder though why I made the fields private. :thinking:

bikeshedder: Heck, I really have to look into that. I created a tracking issue for that:
- #481

bombsimon: Sorry if this grew, especially now with the `AsyncConnectionConfig` on main which isn't ideal. Should we put a pin in this until a more generic approach is reached? If so, should I revert the previous change so releasing main isn't blocked?

One thing with `#[non_exhaustive]` that surfaced with the doc test that I forgot about, it's not possible to build them externally. 

Internally I can write a struct literal like this:

```rust
ManagerConfig {
    connection_timeout: Some(Duration::from_secs(5)),
    response_timeout: None,
    ..Default::default()
}
```

But externally the user would have to write this:

```rust
let mut config = ManagerConfig::default();
config.connection_timeout = Some(Duration::from_secs(5));
config.response_timeout = None;
```

[Reference to why](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html#functional-record-updates).

I personally don't love that pattern so that could be one argument for a builder pattern I guess.

bikeshedder: Ideally you would never need to hardcode this in your application anyways. Overwritten defaults, yes, maybe. But in an actual application you would probably want to have a `VALKEY__CONNECTION_TIMEOUT=2s` environment variable (or some sort of configuration file).

I've written a rationale here:

- [#481](https://github.com/deadpool-rs/deadpool/issues/481#issuecomment-4129875963)

Not being able to spread `..Default::default()` is quite a bummer. I was a bit curious why that's the case and got the answer in the document you linked:

> We’d expect this code to work without the non_exhaustive attribute:
> 
> ```rust
> let c = Config { width: 1920, height: 1080, ..Config::default() };
> println!("{:?}", c);
> ```
> 
> Although outside of the defining crate, it will not, because Config could, in the future, contain private fields that the user didn’t account for.

It's too late for me and I seriously don't care THAT much. I just want the `deadpool-*` crates to be consistent. e.g. in `deadpool-postgres` I didn't even bother adding `#[non_exhaustive]` to the config struct. If I ever need to add a new configuration variable it's going to be a breaking change. But one that's easy to fix. Just bump the version number and see if your code still compiles. If you were matching the config it's probably for the better that it's breaking because you might have missed a config variable that you are now interested in.

If you want to call it the day just make it a plain struct without the `#[non_exhaustive]`. That'd be fine for me as it has the best DX in my opinion.

bikeshedder: fyi. I've spent the past hour debating this with myself ... I see it like that:

- When adding `#[non_exhaustive]` to a config struct a builder pattern should be added for convenience. Though I think the fields should stay public for easier access (and being able to `match`). I despise setters and getters for config structs.
- Leaving out `#[non_exhaustive]` will introduce breaking changes when new fields are added. The structs are fully (de)serializable and adding a fields does change the representation. Maybe it is a good thing that this is a breaking change?

bikeshedder: Let's get this merged. Just drop the `#[non_exhaustive]` so it's on-par with the other `deadpool-*` config structures. I'm fine with that solution. It's good enough. :sweat_smile:

bombsimon: Sounds good! I think just fixing compiler issues and adding defaults on breakage on new fields is reasonable for now. It's likely not bound to happen **that** often.

Really appreciate your time on this and the discussion! I've pushed removal of `#[non_exhaustive]` but am happy to continue discussion here or linked issue.

bikeshedder:
There is an interesting split now. There is a `deadpool_redis::Config` structure which comes with a `create_pool` method. Setting the new connection and request timeouts is only possible if you construct the manager directly and not via the `Config` structure.

I guess the most correct approach here would be to add the timeout fields to the `Config` struct as well and construct the `Manager` via the new `Manager::new_with_config` method.

Also I just now realized that the new `ManagerConfig` doesn't really need to be (de)serializable as it's not used as part of the `Config` struct. It is not wrong to add `serde` support here, too, but it's really not necessary and probably better to remove it so being `Deserialize` and `Serialize` doesn't become public API.

I'm so sorry for mixing that up.  :facepalm:

bombsimon:
> I guess the most correct approach here would be to add the timeout fields to the `Config` struct as well and construct the `Manager` via the new `Manager::new_with_config` method.

The timeout fields are on the `Config` [here](https://github.com/bombsimon/deadpool/blob/6475b019fe731668c1c7229ec35d0fa3e83cb362/crates/deadpool-redis/src/config.rs#L60-L68) and the manager is built with `new_with_config` [here](https://github.com/bombsimon/deadpool/blob/6475b019fe731668c1c7229ec35d0fa3e83cb362/crates/deadpool-redis/src/config.rs#L107-L113).


> Also I just now realized that the new `ManagerConfig` doesn't really need to be (de)serializable as it's not used as part of the `Config` struct. It is not wrong to add `serde` support here, too, but it's really not necessary and probably better to remove it so being `Deserialize` and `Serialize` doesn't become public API.

There's no serialization macro on `ManagerConfig` from what I can see from [here](https://github.com/bombsimon/deadpool/blob/6475b019fe731668c1c7229ec35d0fa3e83cb362/crates/deadpool-redis/src/lib.rs#L166-L167)

Maybe I'm misunderstanding both comments? 😅

PR #477, opened 2026-03-14
fix devcontainer failed to start

### LINKED ISSUE #476: Devcontainer fails to start
Hi, I tried contributing to **deadpool-redis** using GitHub Codespaces, but the devcontainer fails to start.

The issue appears to come from `redis-sentinel` in `docker-compose`, where the referenced image tag no longer exists.

While looking into this, I also noticed the devcontainer build takes a long time because it installs `cargo-release` and `cargo-outdated` via `cargo install`. It might be better to move these to `postCreateCommand` to speed up image builds.

I also noticed that even small changes trigger a full Docker image rebuild. Adding a `.dockerignore` file might help avoid unnecessary rebuilds.

I’ll open a PR with fixes once I’ve cleaned this up.

dreamoutbox:
Fix the devcontainer failing to start in Codespaces and improve build performance.

Changes:

* Fix `redis-sentinel` image tag in `docker-compose`
* Move `cargo-release` and `cargo-outdated` installation to `postCreateCommand`
* Add `.dockerignore` to avoid unnecessary full Docker rebuilds

Closes #476

### DIFF
--- a/.devcontainer/.dockerignore
+++ b/.devcontainer/.dockerignore
@@ -1,0 +1,7 @@
+.env
+compose.yml
+devcontainer.json
+Dockerfile
+postgres.env
+rabbitmq.env
+post-create.sh
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -17,6 +17,3 @@ USER vscode
 
 ARG RUST_VERSION
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain ${RUST_VERSION} --no-modify-path -y
-
-RUN /home/vscode/.cargo/bin/cargo install cargo-release
-RUN /home/vscode/.cargo/bin/cargo install --locked cargo-outdated
--- a/.devcontainer/compose.yml
+++ b/.devcontainer/compose.yml
@@ -51,7 +51,7 @@ services:
 
   redis-sentinel:
     # https://hub.docker.com/r/bitnami/redis-sentinel
-    image: bitnami/redis-sentinel:7.4
+    image: bitnami/redis-sentinel:latest
     environment:
       ALLOW_EMPTY_PASSWORD: yes
     volumes:
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+cargo install cargo-release
+cargo install --locked cargo-outdated
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 cargo install cargo-release
 cargo install --locked cargo-outdated
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,7 @@
   "dockerComposeFile": "compose.yml",
   "service": "app",
   "workspaceFolder": "/workspace",
+  "postCreateCommand": ["/bin/bash", "./.devcontainer/post-create.sh"],
   // Configure tool-specific properties.
   "customizations": {
     // Configure properties specific to VS Code.


.devcontainer/compose.yml
```diff
@@ -51,7 +51,7 @@ services:
 
   redis-sentinel:
     # https://hub.docker.com/r/bitnami/redis-sentinel
-    image: bitnami/redis-sentinel:7.4
+    image: bitnami/redis-sentinel:latest
```
bombsimon, line 54: Note to maintainer: this is related to https://github.com/bitnami/containers/issues/83267 and using `latest` is the recommended way.

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
bombsimon: nit: Since this script doesn't need `bash`, maybe a more portable shebang can be used and drop `/bin/bash` from `postCreateCommand`?


```suggestion
#!/bin/sh
```

E.g. by default `bash` is not installed on Alpine and `sh` is part of the POSIX contract.

bombsimon (approved):
I'm no maintainer nor collaborator but maybe a review will help ease load of maintainer anyway! 

Nice fix! 🎉

bikeshedder:
Awesome. Thanks a lot. I'll look into that shortly. :+1:

.devcontainer/.dockerignore
bikeshedder, line 1: I wonder what's the reason for adding this file. The `Dockefile` doesn't copy anything from the `.devcontainer` directory anyways so this does nothing.

.devcontainer/compose.yml
bikeshedder: I know this is recommended but I'm quite unhappy about this practice. That way things randomly break or can be fixed by a container rebuild.

With a pinned version everyone using the devcontainer will run a known good version of the images.

.devcontainer/Dockerfile
```diff
@@ -17,6 +17,3 @@ USER vscode
 
 ARG RUST_VERSION
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain ${RUST_VERSION} --no-modify-path -y
-
-RUN /home/vscode/.cargo/bin/cargo install cargo-release
-RUN /home/vscode/.cargo/bin/cargo install --locked cargo-outdated
```
bikeshedder, line 22: Why move this to the `post-create` step of the container? I know a lot of people use a quite vanilla container and then add things via features (or post-create scripts) but honestly I never really understood the rationale behind that.

bikeshedder (changes requested):

.devcontainer/.dockerignore
dreamoutbox: You might be right. 

normally the Dockerfile doesn't copy from .devcontainer. but I've noticed Codespaces sometimes rebuild the image when I completely stop and then start it. if there are changes in .devcontainer, even when I haven't changed the Dockerfile. That seemed to cause full rebuilds, so I added the dockerignore to try to prevent that.

.devcontainer/compose.yml
dreamoutbox: We can't avoid this. We need to keep using the bitnami/redis image or switch to the valkey image.
let's decide which one to use later.

.devcontainer/Dockerfile
dreamoutbox: The `cargo install` step greatly increases Docker build time.

I recently saw a Codespace build take 30–40 minutes on a 4‑core, 16GB plan just to make a tiny code change, which feels excessive.

An alternative is to provide a prebuilt image, but it should be hosted from an official location rather than me hosting it personally.

dreamoutbox (commented):
reply to review

.devcontainer/Dockerfile
bikeshedder: I see. The rebuild takes a bit of time on my machine, too, but it's just a couple of minutes. Instead of `cargo install`-ing `cargo-outdated` we could also copy the binary from a docker image. But there are no official builds of that available as far as I'm aware of.

It probably makes sense to move this into a `post-create.sh` script so you can start working sooner even if the `cargo outdated` command isn't available, yet.

PR #478, opened 2026-03-15
make `managed_reexports` macro export `QueueMode`

### LINKED ISSUE #472: `QueueMode` is from `deadpool` crate, not `deadpool-redis`, causing import confusion
When configuring a Redis connection pool with `deadpool-redis`, users need to set the `queue_mode` field in `PoolConfig`. However, the `QueueMode` type is actually defined in the `deadpool::managed` module, not re-exported by `deadpool-redis`. This causes confusion and compilation errors.

**Current situation**:

Users might try to import `QueueMode` from `deadpool-redis`:
```rust
use deadpool_redis::{Config, PoolConfig, QueueMode};  // ❌ Error: QueueMode not found
```

But the correct import should be:
```rust
use deadpool::managed::QueueMode;  // ✅ Works, but requires adding `deadpool` dependency
```

**Example code that fails**:

```rust
use deadpool_redis::{Config, Pool, Runtime, PoolConfig, Timeouts};
use std::env;
use std::time::Duration;

pub async fn create_pool() -> Result<(), Box<dyn std::error::Error>> {
    let url = env::var("REDIS_URL")?;
    let mut cfg = Config::from_url(url);
    
    cfg.pool = Some(PoolConfig {
        max_size: 16,
        timeouts: Timeouts {
            wait: Some(Duration::from_millis(200)),
            create: Some(Duration::from_millis(500)),
            recycle: Some(Duration::from_millis(100)),
        },
        queue_mode: ???, // What should go here? QueueMode is not in scope
    });
    
    let pool = cfg.create_pool(Some(Runtime::Tokio1))?;
    Ok(())
}
```

**Possible solutions**:

1. **Re-export `QueueMode`** in `deadpool-redis` so users don't need to depend on `deadpool` directly:
   ```rust
   // In deadpool-redis lib.rs
   pub use deadpool::managed::QueueMode;
   ```

2. **Add clearer documentation** showing the correct import path and an example of configuring `queue_mode`.

3. **Make `queue_mode` optional** with a sensible default (e.g., `QueueMode::Fifo`) so users don't need to specify it unless they want to change it.

**Current workaround**:

Users must add `deadpool` as a direct dependency and import from there:

```toml
[dependencies]
deadpool-redis = "0.12"
deadpool = "0.10"  # Required for QueueMode
```

```rust
use deadpool::managed::QueueMode;
use deadpool_redis::{Config, PoolConfig, Timeouts};

// Now can use QueueMode::Fifo or QueueMode::Lifo
```

**Expected behavior**:

Either `QueueMode` should be re-exported by `deadpool-redis`, or `PoolConfig` should provide a way to set the queue mode without requiring users to import types from the base `deadpool` crate.

**Environment**:
- `deadpool-redis` version: 0.12.x
- `deadpool` version: 0.10.x
- Rust version: 1.70+

dreamoutbox:
fix #472

### DIFF
--- a/crates/deadpool-redis/tests/redis.rs
+++ b/crates/deadpool-redis/tests/redis.rs
@@ -28,6 +28,14 @@ fn create_pool() -> deadpool_redis::Pool {
 }
 
 #[tokio::test]
+async fn deadpool_reexports_queue_mode() {
+    use deadpool_redis::QueueMode;
+
+    let _ = QueueMode::Fifo;
+    let _ = QueueMode::Lifo;
+}
+
+#[tokio::test]
 async fn test_pipeline() {
     use deadpool_redis::redis::pipe;
     let pool = create_pool();
--- a/crates/deadpool/src/managed/reexports.rs
+++ b/crates/deadpool/src/managed/reexports.rs
@@ -54,5 +54,7 @@ macro_rules! managed_reexports {
         #[doc=concat!("Type alias for using [`deadpool::managed::HookError`] with [`", $crate_name, "`].")]
         pub type HookError = deadpool::managed::HookError<$Error>;
 
+        #[doc=concat!("Type alias for using [`deadpool::managed::QueueMode`] with [`", $crate_name, "`].")]
+        pub type QueueMode = deadpool::managed::QueueMode;
     };
 }


.devcontainer/.dockerignore
```diff
@@ -0,0 +1,7 @@
+.env
+compose.yml
+devcontainer.json
+Dockerfile
+postgres.env
+rabbitmq.env
+post-create.sh
```
bikeshedder: Please remove the [fix devcontainer failed to start](https://github.com/deadpool-rs/deadpool/pull/478/changes/11bd3f67b601267b966b80aa9a3662aaf179ea4c) commit from this PR.

bikeshedder:
I just edited the PR myself and removed the commit.

Thanks a lot for the fix! :heart:

PR #482, opened 2026-04-10
CI: Hash-pin all actions, apply other suggestions from zizmor

woodruffw:
Hello! Apologies for the cold PR.

I'm opening this in my capacity as one of [uv](https://github.com/astral-sh/uv)'s maintainers; we have a set of downstreams (including `deadpool`!) that we depend on, and we'd like to ensure their CI/CD processes are as hermetic and secure as possible (within the limits of GitHub's platform).

To that effect, this PR contains a few different commits that aim to make `deadpool`'s CI more secure. None of these changes fix _vulnerabilities_; they're purely defense-in-depth changes that will make a future [Trivy-style compromise](https://www.wiz.io/blog/trivy-compromised-teampcp-supply-chain-attack) less fruitful for an attacker. 

To summarize:

- I've hash-pinned all of your action dependencies with `pinact run -v`. I've also added a Dependabot config (with a cooldown setting) that will keep your actions (and their hash-pins/version comments up to date). However, if you'd prefer to not have Dependabot, another option here would be to automate these bumps with `pinact` locally yourselves: https://github.com/suzuki-shunsuke/pinact
- I've disabled `actions/checkout`'s default credential-persistence behavior with `persist-credentials: false`, where possible.
- I've dropped your default `permissions` to `{}`, wherever possible. I've also moved all non-empty permissions into their respective jobs rather than setting them at the entire workspace level.

_Most_ of the above was detected automatically with [zizmor](https://docs.zizmor.sh), which you can [integrate](https://docs.zizmor.sh/integrations/#github-actions) into GitHub Actions if you'd like. I've left that out of this PR however, since not every project wants another thing running in CI. But let me know if you'd like it and I'd be happy to send a follow-up PR!

Last but not least, please let me know if there's any other information I can provide. All of the above was 100% human written and reviewed 🙂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    cooldown:
+      default-days: 7
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/deadpool-diesel.yml
+++ b/.github/workflows/deadpool-diesel.yml
@@ -8,8 +8,8 @@ jobs:
     name: Check integration
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@v1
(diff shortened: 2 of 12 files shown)


bikeshedder:
Sorry for taking so long. That change is perfectly fine. I prefer tagged versions, too. :+1:

PR #485, opened 2026-05-28
Replace `num_cpus` with `std::thread::available_parallelism`

CathalMullan:
Makes use of `std::thread::available_parallelism` instead of needing the external `num_cpus::get`. There's some minor differences in behavior between the two, but nothing substantial, both report logical CPU count in practice.

For context, [`tokio`](https://github.com/tokio-rs/tokio/pull/6709) and [`futures`](https://github.com/rust-lang/futures-rs/pull/2946) have made the same switch in the past.

### DIFF
--- a/crates/deadpool/CHANGELOG.md
+++ b/crates/deadpool/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Remove `num_cpus` dependency
+
 ## [0.13.0] - 2026-02-17
 
 - Update `deadpool-runtime` dependency to version 0.3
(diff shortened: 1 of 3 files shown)


bikeshedder:
If `tokio` has made the switch, too, it sounds like a sensible choice.

@PureWhiteWu What are your thoughts on that?

PureWhiteWu (approved):
There's a minor difference between `num_cpu` and `std::thread::available_parallelism` under cgroup environment:
`num_cpus` uses `ceil` and `std::thread::available_parallelism` uses `floor`.

In `deadpool`'s scenario, I think this is acceptable, and can gain benefits of clean dependencies.

bikeshedder:
Thanks a lot for the PR and the Review! :partying_face:

PR #488, opened 2026-06-19
Fix CI generation, resolve all CI issues

CathalMullan:
The recent changes in #482 weren't applied to the CI generation layer. First commit here doesn't change any behavior, just copies the changes to the jsonnet layer.

Second commit resolves all the pre-existing CI issues. Mainly version bumps and feature re-exports.

### DIFF
--- a/.github/workflows/deadpool-diesel.yml
+++ b/.github/workflows/deadpool-diesel.yml
@@ -3,16 +3,15 @@ defaults:
     working-directory: ./crates/deadpool-diesel
 env:
   RUST_BACKTRACE: 1
-permissions: {}
 jobs:
   check-integration:
     name: Check integration
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd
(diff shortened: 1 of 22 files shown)


CathalMullan:
Hmm, CI networking seems flaky for the one failed job. Could try bumping `CARGO_NET_RETRY`.

CathalMullan:
Ah, seems the flaky CI was related to a Cargo bug fixed in 1.96.1: https://blog.rust-lang.org/2026/06/30/Rust-1.96.1

bikeshedder:
I like those tagged versions. :heart:

Is there a good way to check the versions of the used actions so they can be updated more easily?

CathalMullan:
@bikeshedder I've used [`pinact`](https://github.com/suzuki-shunsuke/pinact) on my own projects, which can print a diff compared to the latest.

Only issue here is due to the generated CI approach, it complains about the lack of comments.

```shell
> pinact run --update --diff
.github/workflows/deadpool-diesel.yml:12
-       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd
+       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
...
```

bikeshedder:
I wish there was a better way than generating the actions. GitHub Actions are really a bummer to work with.

